PR #10, opened 2024-05-15
project (data/resource) and building block (data)

henryde:


examples/provider-install-verification/main.tf
```diff
@@ -0,0 +1,41 @@
...
+      source = "meshcloud/meshstack"
+    }
+  }
+}
+
+locals {
+  workspace = "henrys-workspace"
+}
+
+provider "meshstack" {
```
j0g3sc, line 13: probably not critical, but better put secrets here.
guess the whole example can be reworked a bit later.

internal/provider/buildingblock.go
```diff
@@ -1,43 +1,42 @@
...
+	Inputs               []MeshBuildingBlockIO     `json:"inputs" tfsdk:"inputs"`
+	ParentBuildingBlocks []MeshBuildingBlockParent `json:"parentBuildingBlocks" tfsdk:"parent_building_blocks"`
 }
 
 type MeshBuildingBlockIO struct {
-	Key       string `json:"key"`
-	Value     any    `json:"value"`
-	ValueType string `json:"valueType"`
+	Key       string `json:"key" tfsdk:"key"`
+	Value     string `json:"value" tfsdk:"value"`
```
j0g3sc, line 30: we support int and bool as well in the backend.

internal/provider/client.go
```diff
@@ -141,50 +154,213 @@ func (c *MeshStackProviderClient) lookUpEndpoints() error {
...
+	}
+
+	defer res.Body.Close()
+
+	data, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: create will also succeed if the project already exists
```
j0g3sc, line 286: shouldn't be the case anymore

j0g3sc (approved):
minor comments, mainly validated that the schemata look correct to me :+1:

internal/provider/buildingblock.go
henryde: Hm. according to the docs we also have FILE, LIST, SINGLE_SELECT.

PR #18, opened 2024-06-13
feat: projects data source

malhussan:


internal/provider/projects_data_source.go
```diff
@@ -0,0 +1,165 @@
...
+	projects, err := d.client.ReadProjects(workspaceIdentifier, paymentMethodIdentifier)
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to read projects", err.Error())
+		return
+	}
+
+	var state MeshProjectsSchema
+
+	for _, p := range *projects {
+		project := MeshProject{
```
henryde, line 149: d: `projects` is already a list of `MeshProjects`, why create new ones?

henryde (changes requested):

internal/provider/projects_data_source.go
malhussan: agreed, no need for it.

internal/provider/projects.go
```diff
@@ -0,0 +1,12 @@
+package provider
+
+type MeshProjectsSchema struct {
```
henryde, line 3: f: This type belongs primarily to the client and should be called `MeshProjects`

internal/provider/projects.go
```diff
@@ -0,0 +1,12 @@
...
+	WorkspaceIdentifier     string        `tfsdk:"workspace_identifier"`
```
henryde, line 4: f: should all have json annotations as well

henryde (changes requested):

internal/provider/projects.go
malhussan: I've moved it to client as part of a nested response struct
https://github.com/meshcloud/terraform-provider-meshstack/blob/d77af63254ab0f1405fbd14537fe478f3146a31c/internal/provider/client.go#L307

internal/provider/projects_data_source.go
```diff
@@ -0,0 +1,156 @@
...
+		WorkspaceIdentifier     string        `json:"workspaceIdentifier" tfsdk:"workspace_identifier"`
+		PaymentMethodIdentifier *string       `json:"paymentMethodIdentifier" tfsdk:"payment_method_identifier"`
+		Projects                []MeshProject `json:"meshProjects" tfsdk:"projects"`
+	}
+
+	state.Projects = *projects
+	state.WorkspaceIdentifier = workspaceIdentifier
+	state.PaymentMethodIdentifier = paymentMethodIdentifier
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```
henryde, line 155: s: this whole part could be replaced with 3 `resp.State.SetAttribute(...)` making the nested struct unnecessary.

henryde (approved):

PR #25, opened 2024-08-02
fix: Set Accept header when getting a Building Block

sttomm:
meshStack enforces the Accept header soon, so we have to make sure to always provide it


malhussan (approved):
lgtm

PR #38, opened 2025-07-25
Feature/workspace

henryde:


internal/provider/workspace_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+				Validators: []validator.String{
+					stringvalidator.OneOf([]string{"meshWorkspace"}...),
+				},
+			},
+
+			"metadata": schema.SingleNestedAttribute{
+				Required: true,
+				Attributes: map[string]schema.Attribute{
+					"name": schema.StringAttribute{
+						MarkdownDescription: "Name of the workspace. Is used to identify the workspace.",
```
nroi, line 57: d: I know we want the property to be named `name` because we draw inspiration from k8s, but the description reads as if the `name` corresponds to what is shown as the workspace name in the panel (and not the identifier).

If we use this description in other places too and want to be consistent now, I'm fine it, just want to share my perspective, because for someone who's familiar with the panel and not so familiar with k8s, this entire name/identifier business is rather confusing.

nroi: FYI this is how we document it in the API docs 
image

nroi (approved):
lgtm, but I've never written any code in go and have only little familiarity with terraform, so perhaps wait for Jo's review.

malhussan (approved):
lgtm! can you also add an example for workspace import

```hcl
# import via workspace identifier
terraform import 'meshstack_workspace.example' 'my-workspace-identifier'
```

PR #70, opened 2025-11-03
chore: use new API endpoint for login and improve error message in case of login error

j0g3sc:


client/client.go
```diff
@@ -86,20 +91,25 @@ func (c *MeshStackProviderClient) login() error {
...
+	req, _ := http.NewRequest(http.MethodPost, loginPath, bytes.NewBuffer(payload))
+	req.Header.Add("Content-Type", "application/json")
 
 	res, err := c.httpClient.Do(req)
 
 	if err != nil {
 		return err
 	} else if res.StatusCode != 200 {
-		return errors.New(ERROR_AUTHENTICATION_FAILURE)
+		return errors.New(fmt.Sprintf("Status %d: %s", res.StatusCode, ERROR_AUTHENTICATION_FAILURE))
```
nroi: d: we're still using `ERROR_AUTHENTICATION_FAILURE` for all status codes other than 200?

PR #74, opened 2025-11-19
feat: added payment method resource and data endpoint

younGihan:
added terraform provider endpoint for payment methods including resource creation, deletion and data reads

### DIFF
--- a/examples/resources/meshstack_payment_method/resource.tf
+++ b/examples/resources/meshstack_payment_method/resource.tf
@@ -16,7 +16,7 @@ resource "meshstack_payment_method" "example" {
     amount          = 10000
     tags = {
       CostCenter = ["0000"]
-      Type    = ["production"]
+      Type       = ["production"]
     }
   }
 }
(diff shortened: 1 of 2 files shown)


examples/resources/meshstack_payment_method/resource.tf
```diff
@@ -0,0 +1,22 @@
...
+
+resource "meshstack_payment_method" "example" {
+  metadata = {
+    name               = "my-payment-method"
+    owned_by_workspace = data.meshstack_workspace.example.metadata.name
+  }
+
+  spec = {
+    display_name    = "My Payment Method"
+    expiration_date = "2025-12-31T23:59:59Z"
```
nroi: f: I tried it locally with this expiration date, and it didn't work. looks like you need specify the date only:
```suggestion
    expiration_date = "2025-12-31"
```

Image

nroi (approved):
left a few comments and also found a bug unfortunately, but perhaps it's a valid approach to merge it now and fix the API at a later point? we can also discuss this via slack.

PR #151, opened 2026-04-17
feat: meshstack_building_block resource (building block v3)

grubmeshi:
Adds the `meshstack_building_block` resource (v3, preview API) for managing Building Blocks via the meshObject API.

## What it delivers

- Generic, typed inputs via `jsonencode(...)`; sensitive inputs via a `sensitive` block with `secret_version`-controlled rotation.
- In-place version upgrade by changing `building_block_definition_version_ref.uuid`.
- `moved` support from `meshstack_buildingblock` (v1) and `meshstack_building_block_v2`.
- Status surfaced — `status.latest_run_uuid` (latest modifying run) and the new computed `status.latest_dry_run_uuid`; runs awaited with run-log diagnostics on failure, with a configurable `timeouts` block (create/update/delete).

## Behavior contract

- **Replacement:** only `target_ref.*` forces replacement (you cannot move a Building Block to another tenant/workspace). `display_name` and `inputs` update in place. `parent_building_blocks` change only as part of a version upgrade — changing them on their own forces replacement.
- **Reruns:** a run is triggered on a version change, an input change, a sensitive-input rotation (`secret_version` bump), a `content_hash` change, or a parent change carried by a version upgrade; a rename or a no-op does not.
- **Partial / shared input ownership:** a configuration manages only the inputs it declares in `spec.inputs`; an input set by another party but omitted is preserved server-side and surfaced read-only in `all_inputs`, not as drift. This lets a platform operator manage operator inputs while the consumer owns user inputs, each declaring only its own.
- **Upgrades:** only to the latest released version of the same definition.
- **Sensitive inputs:** use the `sensitive` block iff the definition declares the input sensitive; a `moved` migration warns to re-supply the secret (it cannot travel through state).
- A version change on a non-completed block, and operator inputs set by a non-operator key, surface clear errors instead of raw conflicts or silent drift.


client/building_block_v2.go
```diff
@@ -0,0 +1,234 @@
...
+	Lifecycle  MeshBuildingBlockV2Lifecycle       `json:"lifecycle" tfsdk:"-"`
+	// LatestRunUuid is nil if permissions don't allow reading the run (e.g. because run_transparency is false).
+	LatestRunUuid *string `json:"latestRunUuid" tfsdk:"latest_run_uuid"`
+}
+
+type MeshBuildingBlockOutput struct {
+	Value          types.Any                                        `json:"value" tfsdk:"value"`
+	ValueType      enum.Entry[MeshBuildingBlockIOType]              `json:"valueType" tfsdk:"value_type"`
+	AssignmentType enum.Entry[MeshBuildingBlockInputAssignmentType] `json:"assignmentType" tfsdk:"assignment_type"`
+}
```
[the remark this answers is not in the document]
grubmeshi: indeed, also check resource schema definition is this is wrongly set up?

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,981 @@
...
+							},
+							"uuid": schema.StringAttribute{
+								MarkdownDescription: "UUID of the target tenant. Required when `kind = \"meshTenant\"`, must be omitted for `kind = \"meshWorkspace\"`.",
+								Optional:            true,
+								Validators: []validator.String{
+									stringvalidator.ExactlyOneOf(
+										path.MatchRelative().AtParent().AtName("uuid"),
+										path.MatchRelative().AtParent().AtName("name"),
+									),
+									// P8 step 3: meshTenant requires uuid, meshWorkspace requires name.
```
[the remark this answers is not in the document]
grubmeshi: Yes, indeed.

internal/provider/acctest/testconfig/build_building_block.go
```diff
@@ -0,0 +1,53 @@
+package testconfig
+
+import (
+	"testing"
+)
+
+func BBWorkspace(t *testing.T) (config Config, buildingBlockAddr Traversal, buildingBlockDefinitionAddr Traversal) {
```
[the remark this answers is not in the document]
grubmeshi: Keeping the `BB`/`BBD` prefixes: they're an established convention across the sibling testconfig builders (`BBv2Workspace`, `BBv1Tenant`, `BBDManual`, …), and `Workspace`/`Tenant` are already taken by the meshObject builders, so renaming only these two would be inconsistent. Resolving as won't-fix.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,859 @@
...
+	stateSpec := generic.GetAttribute[client.MeshBuildingBlockV2Spec](ctx, req.State, path.Root("spec"), &resp.Diagnostics, converterOptions...)
+	planSpec := generic.GetAttribute[client.MeshBuildingBlockV2Spec](ctx, req.Plan, path.Root("spec"), &resp.Diagnostics, converterOptions...)
+
+	if planSpec.BuildingBlockDefinitionVersionRef.Uuid != stateSpec.BuildingBlockDefinitionVersionRef.Uuid {
+		retriggerRun()
+	} else if planContentHash := planSpec.BuildingBlockDefinitionVersionRef.ContentHash; planContentHash != nil {
+		if stateContentHash := stateSpec.BuildingBlockDefinitionVersionRef.ContentHash; stateContentHash == nil || *stateContentHash != *planContentHash {
+			retriggerRun()
+		}
+	}
```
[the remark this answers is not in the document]
grubmeshi: ModifyPlan and Update now share the same `rerunNeeded(plan, state)` predicate, so they no longer diverge on content_hash handling. Removing a previously-set content_hash is intentionally not a rerun (documented in `rerunNeeded`).

client/building_block_v2.go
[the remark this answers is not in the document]
grubmeshi: Fixed: the field now uses `MeshBuildingBlockDefinitionOutputAssignmentType`. The resource schema was already correct (it uses `MeshBuildingBlockDefinitionOutputAssignmentTypes.Markdown()`), so only the Go client struct needed the type change.

client/building_block_v2_test.go
```diff
@@ -0,0 +1,173 @@
...
+		UserAgent: "test",
+	}
+
+	_, err = internal.DoRequest[MeshBuildingBlockV2](context.Background(), c, http.MethodGet, rootUrl)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "expected response body")
+}
+
+// TestDoRequest_HtmlErrorBodyReturnsCleanHttpError verifies that a non-2xx HTML response
+// produces a clean HttpError (with raw body) rather than joining an unmarshal error (P3.2).
```
grubmeshi: f: scan codebase for leftover review markers such as P3.2, remove them entirely.

client/building_block_v2_test.go
```diff
@@ -0,0 +1,173 @@
...
+			if tt.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestMeshBuildingBlockV2_CreateSuccessful_NilUuidNoPanic(t *testing.T) {
```
grubmeshi: f: turn this into parameterized test similar to DeleteSuccessful covering all cases

client/building_block_v2.go
```diff
@@ -0,0 +1,240 @@
...
+	"fmt"
+	"slices"
+
+	"github.com/meshcloud/terraform-provider-meshstack/client/internal"
+	"github.com/meshcloud/terraform-provider-meshstack/client/types"
+	"github.com/meshcloud/terraform-provider-meshstack/client/types/enum"
+)
+
+const (
+	BUILDING_BLOCK_LIFECYCLE_STATE_ACTIVE              = "ACTIVE"
```
grubmeshi: f: also turn this into an enum.Enum and use camel case vars (see BuildingBlockStatus below for example)

examples/resources/meshstack_building_block/resource_03_operator_inputs.tf
```diff
@@ -0,0 +1,20 @@
+resource "meshstack_building_block" "example_operator" {
+  spec = {
+    building_block_definition_version_ref = one(data.meshstack_building_block_definitions.example.building_block_definitions).version_latest
+
+    display_name = "my-operator-building-block"
+    target_ref   = data.meshstack_workspace.example.ref
+
+    inputs = {
```
grubmeshi: f: instead of adding another example which is quite similar to the 01_workspace one, remove this example and the acceptance tests has just a different BBD using the same input keys as the 01_workspace, making the 01_workspace example 1:1 reusable (but with changed behavior by BBD like non updateable or op_config). it would be fine if that requires more explicit work in the acc test subtest and less testconfig builders. testconfig builders are supposed to be reused, and the acc tests are rather specific for BB resource.

examples/resources/meshstack_building_block/resource_02_tenant.tf
```diff
@@ -0,0 +1,24 @@
...
+  spec = {
+    # Alternatively, use version_latest_release to target only released versions
+    building_block_definition_version_ref = one(data.meshstack_building_block_definitions.example.building_block_definitions).version_latest
+
+    display_name = "my-tenant-building-block"
+    target_ref   = one(data.meshstack_tenants.example.tenants).ref
+
+    inputs = {
+      name = {
+        value = jsonencode("my-name")
```
grubmeshi, line 11: f: remove size/environment inputs, but instead replace it with a sensitive user input like, key "sensitive_user_input" and explain succinctly how this works with secret_version or ephemeral input sources. Note that the resource docs already automatically have a section "Handling Write-Only Attributes (Secrets)" which can be referred to in this short comment. this has an implication to the acc tests and check if other subtests become obsolete if this tenant example already shows how sensitive user inputs work. of course, a dedicated BBD must be built/adapted which can be inplace built (as test support file), without touching too much testconfig builders code if possible.

examples/resources/meshstack_building_block/resource_07_consumer_non_updateable.tf
```diff
@@ -0,0 +1,19 @@
+resource "meshstack_building_block" "example_consumer" {
+  depends_on = []
+
+  spec = {
+    building_block_definition_version_ref = one(data.meshstack_building_block_definitions.example.building_block_definitions).version_latest
+
+    display_name = "my-consumer-building-block"
+    target_ref   = data.meshstack_workspace.example.ref
+
+    inputs = {
```
grubmeshi: f: same idea as for 03_operator inputs, the could would be that documented examples show tenant/workspace BBs with sensitive inputs, and everything else is part of acc test for now (test support files)

examples/resources/meshstack_building_block/test-support_moved_from_v1.tf
```diff
@@ -0,0 +1,4 @@
+moved {
+  from = meshstack_buildingblock.placeholder
```
grubmeshi: f: check if placeholder address is actually known which could be if the actual resource address doesn't have a random suffix to avoid resource name collisions. this would simplify loading this testsupport file as no extra "testconfig.Set" adaptation is required then. also check other test support files for similar placeholder removal simplification potential.

examples/resources/meshstack_building_block/test-support_04_sensitive_user_input.tf
```diff
@@ -0,0 +1,22 @@
+resource "meshstack_building_block" "sensitive_user_input" {
```
grubmeshi, line 1: see comment in resource_02_tenant.tf, i think that test support could then be removed if sensitive inputs are already shown to work there. also this file, once removed, looses all placeholder problems (omit this when analyzing other placeholder removal optimiation)

examples/resources/meshstack_building_block/test-support_04_sensitive_user_input_bbd.tf
```diff
@@ -0,0 +1,57 @@
+resource "meshstack_building_block_definition" "sensitive_user_input" {
```
grubmeshi, line 1: f: this can be removed as well, see other comment in related files here

internal/provider/service_instance_data_source.go
```diff
@@ -159,3 +159,37 @@ func withValueFromConverterForClientTypeAny() generic.ConverterOption {
...
+// withValueFromConverterForClientTypeAny. Resources (unlike data sources) read
+// their own state back into the Go model via generic.Get, so the clientTypes.Any
+// output value (an interface{}) must be convertible in this direction too —
+// without it the generic walker panics on the unsupported interface kind.
+//
+// It must match the field type *exactly* (mirroring the From converter's
+// in.Type() == clientTypeAny guard) rather than via generic.WithValueToConverterFor,
+// because clientTypes.Any is an interface and every type is AssignableTo it — a
+// WithValueToConverterFor[clientTypes.Any] would greedily match the root model.
+func withValueToConverterForClientTypeAny() generic.ConverterOption {
```
grubmeshi: f: as this change is actually not related to the service instance data source (it just happens to have also an "any" / unstructured client DTO part), move this out into a separate go file (same package) including the sibling method withValueFromConverterForClientTypeAny. revisit the godoc comments and make them accurate and succinct as well.

internal/types/secret/secret.go
```diff
@@ -22,6 +22,10 @@ type Secret struct {
 	Hash    *string `tfsdk:"secret_hash"`
 }
 
+type HashOnly struct {
```
grubmeshi, line 31: add godoc here and to Secret above explaining the idea behind it. Secret is sth which is managed by the resource, and HashOnly is for data source or for resources having read-only attributes. do not reference specific examples, stay general.

internal/types/secret/secret_schema_resource.go
```diff
@@ -48,3 +48,17 @@ func ResourceSchema(opts ResourceSchemaOptions) (result schema.SingleNestedAttri
 		},
 	}
 }
+
+func ReadOnlyResourceSchema(opts ResourceSchemaOptions) (result schema.SingleNestedAttribute) {
```
grubmeshi, line 55: f: add a succinct godoc

internal/validators/discriminated_attributes_validator.go
```diff
@@ -0,0 +1,78 @@
...
+var _ validator.Object = DiscriminatedAttributesValidator{}
+
+// DiscriminatedAttributesValidator validates a discriminated-union object: the string value of
+// the Discriminator attribute selects which one of the controlled attributes must be set, and
+// requires every other controlled attribute to be null.
+//
+// The controlled set is the set of attribute names in RequiredFor's values. Discriminator values
+// not present in RequiredFor are ignored (validate the discriminator itself with
+// stringvalidator.OneOf). Null/unknown objects and unknown attribute values are skipped.
+type DiscriminatedAttributesValidator struct {
```
grubmeshi, line 21: d: analyze if this validator is useful in other places. right now it has one usage and i'm considering using it at other places if possible. if you spot that slight extension/generilzation would make this reusable, propose to do so.

internal/provider/buildingblock_resource.go
```diff
@@ -23,31 +23,32 @@ import (
...
 var (
-	_ resource.Resource              = &buildingBlockResource{}
-	_ resource.ResourceWithConfigure = &buildingBlockResource{}
+	_ resource.Resource              = &buildingblockResource{}
+	_ resource.ResourceWithConfigure = &buildingblockResource{}
 )
 
-func NewBuildingBlockResource() resource.Resource {
-	return &buildingBlockResource{}
+func NewBuildingblockResource() resource.Resource {
```
grubmeshi, line 34: f: add short godoc explaining why its `Buildingblock` and not `BuildingBlock` here.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+			testconfig.Descend("spec", "building_block_definition_version_ref")(testconfig.SetRawExpr(`{ uuid = %s }`, buildingBlockDefinitionAddr.Join("version_latest", "uuid"))),
+			testconfig.Descend("spec", "target_ref")(testconfig.SetAddr(workspaceAddr, "ref")),
+			testconfig.Descend("purge_on_delete")(testconfig.SetRawExpr("true")),
+		).Join(workspaceConfig, buildingBlockDefinitionConfig)
+
+		var bbUuid string
+		ApplyAndTest(t, resource.TestCase{
+			// Runs after teardown (which purges the block): the purged block must be hard-gone.
+			CheckDestroy: func(*terraform.State) error {
+				bb, err := acceptanceClient(t).BuildingBlockV2.Read(context.Background(), bbUuid)
```
grubmeshi, line 1078: f: this reads strange. BB are soft deleted even if purged, they should still exist in the backend. check the meshfed-release public API implementation why the GET endpoint (which Read uses here) returns a nil BB and not a BB object with lifecycle status DELETED (or purged, don't know how it's implemented in backend)

internal/provider/provider_test.go
```diff
@@ -89,6 +91,33 @@ func ApplyAndTest(t *testing.T, testCase resource.TestCase) {
 	resource.Test(t, testCase)
 }
 
+// acceptanceClient builds a real meshStack API client from the test env vars.
+// Only valid in acceptance mode (TF_ACC set); callers must guard with IsMockClientTest.
+func acceptanceClient(t *testing.T) client.Client {
```
grubmeshi: f: move this helper method and along with awaitBBV1succeed below close to where it's used. consider inlining acceptanceClient (it's used there separately but maybe that can be done more elegantly)

internal/clientmock/mock_building_block_v2.go
```diff
@@ -0,0 +1,218 @@
...
+		panic(fmt.Sprintf("deepCopyBB: unmarshal failed: %v", err))
+	}
+	return &cp
+}
+
+// hashSensitiveInputs simulates what the real meshStack API does: converts any submitted
+// sensitive plaintext value into a sha256 hash and removes the plaintext from the stored input.
+// This lets mock-based tests verify that sensitive inputs surface as hashes in state.
+// It operates on the provided map in place; callers must pass a copy (via deepCopyBB).
+func hashSensitiveInputs(inputs map[string]client.MeshBuildingBlockInput) {
```
grubmeshi: f: replace this with backendSecretBehavior which should work out of the box (if not, investigate why). see if skills / instructions here should be improved to note this concerning mock client in acc tests and secrets / sensitive values behavior.

internal/clientmock/mock_building_block_v2.go
```diff
@@ -0,0 +1,218 @@
...
+	return deepCopyBB(stored), nil
+}
+
+func (m MeshBuildingBlockV2Client) Delete(_ context.Context, bbUuid string, purge bool) error {
+	m.Store.Delete(bbUuid)
+	return nil
+}
+
+// SoftDelete marks the building block as lifecycle-DELETED without removing it from the store.
+// This simulates meshStack >= v2026.20.0 soft-delete behaviour for use in P7 tests.
```
grubmeshi: f: remove this unused method and validate wholisitcally that soft delete behavior is addressed as of now. don't test old behavior anymore. also, P7 is a review marker, which would go as well (see other comment asking to remove all dangling review markers).

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+var (
+	_ resource.Resource                = &buildingBlockResource{}
+	_ resource.ResourceWithConfigure   = &buildingBlockResource{}
+	_ resource.ResourceWithImportState = &buildingBlockResource{}
+	_ resource.ResourceWithModifyPlan  = &buildingBlockResource{}
+	_ resource.ResourceWithMoveState   = &buildingBlockResource{}
+)
+
+// buildingBlockPollTimeout is the maximum time to wait for a building block run to complete.
+const buildingBlockPollTimeout = 30 * time.Minute
```
grubmeshi: f: this should not be hardcoded, but reuse the value configured for the resource for update/create (depending on what action is done) from the timeouts config block, see https://oneuptime.com/blog/post/2026-03-20-resource-timeouts-opentofu/view

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+		return
+	}
+	logs, err := r.BuildingBlockRunClient.GetLogs(ctx, *bb.Status.LatestRunUuid)
+	if err != nil {
+		diags.AddWarning("Could not fetch run logs", err.Error())
+		return
+	}
+	for _, step := range logs.Steps {
+		msg := step.DisplayName + " [" + step.Status + "]"
+		if step.UserMessage != nil {
```
grubmeshi: f: also include system message. ensure to display everhting the api returns.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+) {
+	if pollErr != nil {
+		diags.AddError(summary, pollErr.Error())
+	}
+	if bb == nil || bb.Status == nil || bb.Status.LatestRunUuid == nil {
+		return
+	}
+	logs, err := r.BuildingBlockRunClient.GetLogs(ctx, *bb.Status.LatestRunUuid)
+	if err != nil {
+		diags.AddWarning("Could not fetch run logs", err.Error())
```
grubmeshi: f: mention that fetching logs might be forbidden (no run_transparency) or insufficient permissions, validate public API backend behavior as well. it might be that a app team user with a BBD run_transparence false can still fetch logs but only sees the steps. consider this when construction the warning messages for each step below, only telling what steps have failed. also only report the first step failed, as subsequent ones usually fail because previous did not succeed.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+		if step.UserMessage != nil {
+			msg += ": " + truncate(*step.UserMessage, 2000)
+		}
+		diags.AddWarning("Run step: "+step.DisplayName, msg)
+	}
+}
+
+// truncate cuts s to at most limit runes (not bytes), appending "…" if truncated.
+// P13 item 10: slice by runes to avoid splitting multi-byte UTF-8 sequences.
+func truncate(s string, limit int) string {
```
grubmeshi: f: inline into addFailureDiags func below as as truncate := func(...) {} local var

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+		resp.Diagnostics.AddError("Error creating building block", err.Error())
+		return
+	}
+	plan.SetFromClientDto(created, &resp.Diagnostics)
+	resp.Diagnostics.Append(generic.Set(ctx, &resp.State, plan, converterOptions...)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// P13 item 1: short-circuit the poll if the POST response already satisfies CreateSuccessful
```
grubmeshi: f: P13 is a now obsolete / dangling review marker. scan for them and remove all of them. make the PR changes look like reviews have never happened and foxus on the final result.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1109 @@
...
+		return
+	}
+
+	// P4 step 3: compute rerun condition BEFORE calling the client.
+	// C1/P1: also rerun when a sensitive input was rotated (secret_version changed) — invisible to
+	// rerunNeeded since the converted spec hides it. Same helper as ModifyPlan, so plan and apply agree.
+	needsRun := rerunNeeded(plan.Spec, state.Spec) ||
+		secret.AnyVersionChangedIn(ctx, req.Plan.Raw, req.Plan, req.State, &resp.Diagnostics)
+
+	// B2/X2: a version-change PUT is only accepted by the backend when the block is in a completed state
```
grubmeshi: f: again, B2/X2 review markers must go (See other commens)

internal/provider/building_block_v2_data_source.go
```diff
@@ -198,17 +192,29 @@ func (d *buildingBlockV2DataSource) Read(ctx context.Context, req datasource.Rea
...
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("status").AtName("lifecycle"), bb.Status.Lifecycle)...)
-
-	// Read outputs
-	outputs := make(map[string]buildingBlockOutputModel)
-	for key, output := range bb.Status.Outputs {
-		outputs[key] = toResourceModelV2Output(key, output, &resp.Diagnostics)
-	}
-	if resp.Diagnostics.HasError() {
-		return
+	if bb.Status == nil {
```
grubmeshi: f: diligently check if bb.Status actually can be nil. i'm pretty sure in response objects, it's never nil, so no nil check required. ensure this is reflected in the instructions / skills so future reviews flag this. this is a very general approach that client DTO go structs are reused as response and request payloads, making response-only fields like Status poitners to avoid sending them in requests. carefully search also for other places where should nil checks could be removed again if they were added by this PR changes.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+	t.Parallel()
+
+	// 01_workspace_lifecycle walks a workspace-targeted building block through its full life: create,
+	// import, in-place updates (display_name, an input value, content_hash), and finally a parent
+	// change that must force a replace.
+	//
+	// What to watch:
+	//   - The empty-plan step (step 3) is acceptance-only: the in-memory mock does not materialize the
+	//     null input rows the real backend echoes for unconfigured optional inputs, so the drift
+	//     suppression it checks is only observable against a real meshStack.
```
grubmeshi: f: i think this is true, but try to still run it in mock as it does not harm. this reduces drift between mock / non-mock steps. also, look at uses of skipInMock and skipInAcc and thin about where this is really necessary.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+// The subtests below are scenario flows rather than one-assertion-per-case tests: each walks a
+// building block through a sequence of steps. Several scenarios mix steps that run against the
+// in-memory mock with steps that need the real backend; the latter carry SkipFunc: skipInMock, and
+// where a both-modes step has assertions that only hold against the real backend, the
+// ConfigStateChecks are gated on IsMockClientTest().
+
+// skipInMock skips a TestStep when running against the in-memory mock (TF_ACC unset). Use it for
+// steps that need the real meshStack backend inside an otherwise mock-runnable scenario; a skipped
+// step's ConfigStateChecks do not run, and the next step diffs against the last applied state.
+func skipInMock() (bool, error) { return IsMockClientTest(), nil }
```
grubmeshi: skipInMock and skinInAcc are test infra and should be put below IsMockClientTest func (provider_test.go file I think)

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+	// 04_tenant_moved_from_v1 guards the v1→v3 migration: a `moved` block from the legacy
+	// meshstack_buildingblock (v1) to v3 must plan as an in-place Update, never a destroy+recreate
+	// (recreating a live tenant BB is destructive). moveFromV1 leaves target_ref/version_ref unset so
+	// the post-move refresh-Read fills them from the live DTO before the RequiresReplace modifiers
+	// evaluate.
+	//
+	// What to watch: the v1 resource has no wait_for_completion, so its run is still PENDING after
+	// create; the move is an Update and updating a non-terminal block hits the backend's
+	// completed-state guard (409 conflict). The PreConfig awaits SUCCEEDED first — acceptance-only,
+	// since the mock has no such run.
```
grubmeshi, line 356: f: again, this has potential that if could also just check in mock mode as well (where this is a noop). the caveat here can stay though.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+		v3Config := testconfig.Resource{Name: "building_block", Suffix: "_02_tenant"}.Config(t).WithFirstBlock(
+			testconfig.ExtractAddress(&v3Addr),
+			testconfig.Descend("spec", "building_block_definition_version_ref")(testconfig.SetAddr(buildingBlockDefinitionAddr, "version_latest")),
+			testconfig.Descend("spec", "target_ref")(testconfig.SetAddr(tenantAddr, "ref")),
+		).Join(workspaceConfig, projectConfig, platformConfig, landingZoneConfig, tenantConfig, buildingBlockDefinitionConfig)
+
+		movedConfig := testconfig.Resource{Name: "building_block"}.TestSupportConfig(t, "_moved_from_v1").WithFirstBlock(
+			testconfig.Descend("from")(testconfig.SetAddr(v1Addr)),
+			testconfig.Descend("to")(testconfig.SetAddr(v3Addr)),
+		)
```
grubmeshi: note: this is one of the places which could be simplified by the placeholder optimization mentioned in comments in test support files for moved blocks.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+	// A MANAGED_BUILDINGBLOCK_SAVE key is still minted here (workspace A, the definition owner): in mock
+	// mode this proves the provider accepts MANAGED_BUILDINGBLOCK_SAVE in the api_key permission set
+	// (schema validation against client.Permissions.AllCodes() runs in both modes). A MANAGED key may NOT
+	// create a building block cross-workspace — its cross-workspace power is limited to setting operator
+	// inputs via PUT on an existing block. That MANAGED capability and the create-vs-update authorization
+	// boundary are covered on the backend by MeshBuildingBlockManagedSaveScenarios; 07_non_updateable_rejected
+	// is the companion check for a plain workspace-scoped key creating in its own workspace.
+	//
+	// Runs in BOTH modes — deliberately no skipInMock; the mock hands all provider aliases one shared
+	// store and ignores credentials, so it does not enforce permission boundaries.
```
grubmeshi: note: this is the approach I like! although mock might not cover it, the tests should be defined without many skipInMock / skipInAcc exceptions.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue(buildingBlockAddr.String(), tfjsonpath.New("spec").AtMapKey("inputs").AtMapKey("name").AtMapKey("value"), knownvalue.StringExact(`"updated-name"`)),
+						statecheck.ExpectKnownValue(buildingBlockAddr.String(), tfjsonpath.New("all_inputs").AtMapKey("op_config").AtMapKey("value"), knownvalue.StringExact(`"operator-value"`)),
+					},
+				},
+			},
+		})
+	})
+
+	// 06_sensitive_inputs_and_upgrade is the sensitive-input scenario. One shared BBD declares a
```
grubmeshi, line 674: that is a very useful scenario which should be kept, see if that can be exercised combined with the 02_tenant example, where i commented elsewhere that it should showcase using a sensitive user input already. if that would make it quite hard to understand later, propose not doing it.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+					},
+				},
+			},
+		})
+	})
+
+	// 07_non_updateable_rejected is a cross-workspace permission test: a consumer in another workspace
+	// (using a scoped API key) must be rejected when it tries to change an input the BBD marks
+	// non-updateable-by-consumer. Acceptance-only — it needs real permission boundaries and a second
+	// provider alias, which the mock cannot model.
```
grubmeshi, line 881: f: again consider also running in mock although the test is only meaningful against real backend (but could still be green in mock mode)

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+
+	// 08_validation_rejects collects the plan-time and create-time rejection checks into one scenario
+	// over the workspace BBWorkspace config: a target_ref whose kind and identifier disagree, a
+	// malformed version uuid, and a STATIC BBD input wrongly assigned as a customer input.
+	//
+	// What to watch: the target_ref kind/identifier mismatch checks (steps 1-2) are provider-side
+	// validators and run in both modes; the malformed-uuid 400 (step 3) and the STATIC-input
+	// rejection (steps 4-5) need the real backend and carry SkipFunc: skipInMock. The ExpectError
+	// steps establish no state, so ordering them before the valid create in step 4 is safe; in mock
+	// only steps 1-2 run.
```
grubmeshi, line 981: note: this is fine concerning mock/non-mock, as i don't want to bloat the mock client with simulating error conditions too much.

internal/provider/building_block_resource_test.go
```diff
@@ -0,0 +1,939 @@
...
+						})),
+						statecheck.ExpectKnownValue(buildingBlockAddr.String(), tfjsonpath.New("purge_on_delete"), knownvalue.Bool(true)),
+					},
+				},
+			},
+		})
+	})
+}
+
+func bbv3StateChecks(buildingBlockAddr testconfig.Traversal, displayName string) []statecheck.StateCheck {
```
grubmeshi: note: as I've suggested to change the examples tf code, this method might become obsolete and checks must be inlined or simplified. i would be ok if diligent checks are in place for 01_workspace and 02_tenant only, in one of the first test steps, and then checks focus on the actual critical things the further steps try to cover.

grubmeshi:
Superseded by #208 — the branch was renamed to `feature/building-block-v3` (aligning with the meshfed-release branch so its `terraform-provider-acceptance` CI job picks up this code) and rebased onto #207. Same commits/history continue there.

PR #160, opened 2026-05-04
feat: Removes Azure Blueprint and OpenShift template support

tfelix:
These functionalities are deprecated and not used in the platform anymore. They are full removed from the backend and not supported anymore via the API.


grubmeshi (approved):
LGTM, but please get the CI green (regenerate docs).

PR #164, opened 2026-05-06
feat: adapt building_block_v2 to moved createdOn field in upstream API

nroi:
The meshStack API has moved the building block creation timestamp from
metadata.createdOn to status.lifecycle.createdOn. This change updates the
Terraform provider to reflect the new API structure by:

Note: Existing Terraform state will need a refresh after upgrading the
provider to migrate from the old path to the new one.

### DIFF
--- a/client/buildingblock_v2.go
+++ b/client/buildingblock_v2.go
@@ -24,11 +24,8 @@ type MeshBuildingBlockV2 struct {
 }
 
 type MeshBuildingBlockV2Metadata struct {
-	Uuid                string  `json:"uuid" tfsdk:"uuid"`
-	OwnedByWorkspace    string  `json:"ownedByWorkspace" tfsdk:"owned_by_workspace"`
-	CreatedOn           string  `json:"createdOn" tfsdk:"created_on"`
-	MarkedForDeletionOn *string `json:"markedForDeletionOn" tfsdk:"marked_for_deletion_on"`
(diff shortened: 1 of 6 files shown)


client/buildingblock_v2.go
```diff
@@ -54,10 +51,18 @@ type MeshBuildingBlockV2Create struct {
 	Spec MeshBuildingBlockV2Spec `json:"spec" tfsdk:"spec"`
 }
 
+type MeshBuildingBlockV2Lifecycle struct {
+	State               string  `json:"state" tfsdk:"state"`
+	CreatedOn           string  `json:"createdOn" tfsdk:"created_on"`
```
grubmeshi, line 56: Just FYI, as we've had trouble with dates (different nanosecond precision), I tend to not expose those datetime fields anymore in TF resources. Things like `Lifecycle.State` can stay, though it's a bit awkward as if a managed resource gets deleted, it will be removed from tf state anyhow.

For this PR, your change looks fine. Just the BBv3 resource should not have such brittle datetime fields or even leave out the lifecycle fields entirely.

Thinking about it again, how about only removing the fields above and not adding any lifecycle fields at all?

grubmeshi (approved):

client/buildingblock_v2.go
nroi: I don't want to tackle too many things with a single PR. The scope of this PR to only make the tf provider compatible to upstream changes. And I'm not sure which side effects it can introduce to remove it now.

PR #168, opened 2026-05-13
fix: use types.List instead of []types.String for options fields

tob1aswe1ss:
The tagValueSingleSelect and tagValueMultiSelect model structs used []types.String for their Options and DefaultValue fields. This caused "Value Conversion Error" when these fields received values not fully resolved at plan time (e.g., variable references), because the Terraform Plugin Framework cannot handle Go slices with unknown elements.

Changed to types.List which properly supports unknown/null semantics during the plan phase.

BD-2435

## Summary by CodeRabbit

* **New Features**
  * Tag definitions can derive single-select and multi-select options from Terraform locals for more flexible tag configs.
* **Bug Fixes**
  * Improved handling of null/unknown lists during validation and default assignment; numeric option values are coerced to strings and invalid objects are rejected.
* **Tests**
  * Added acceptance tests covering local-sourced options, numeric coercion, and rejection of invalid option elements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/meshcloud/terraform-provider-meshstack/pull/168?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/provider/tag_definition_resource.go
+++ b/internal/provider/tag_definition_resource.go
@@ -18,7 +18,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 
 	"github.com/meshcloud/terraform-provider-meshstack/client"
 	"github.com/meshcloud/terraform-provider-meshstack/internal/modifiers/tagdefinitionmodifier"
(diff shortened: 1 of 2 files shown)


JohannesRudolph:
## Review: `types.List` migration for options fields

### State Compatibility ✅

No risk. The `schema.ListAttribute{ElementType: types.StringType}` declaration in `Schema()` is **unchanged** — only the Go struct fields were updated (`[]types.String` → `types.List`). Terraform state is stored and validated against the schema, not the Go struct types, so existing state will not produce any plan diff after upgrading.

### Type Safety ✅

`types.List` with `ElementType: types.StringType` is **equally strict** as `[]types.String` was. The element type is enforced at the schema level by Terraform, not by the Go struct.

One nuance worth knowing: Terraform's type system **automatically coerces primitive literals** (numbers, booleans) to strings in a `list(string)` context — e.g. `options = [1, 2, 3]` is accepted as `["1", "2", "3"]`. This was true before this PR as well (the coercion happens at the HCL evaluation layer, before provider code is reached).

### Tests Added

Added `TestTagDefinitionResource_OptionsRejectObjectElements` which verifies that a genuinely incompatible element type (an object — `[{ not_a_string = true }]`) is rejected with a type error. The test comment also documents the number coercion behaviour so the distinction is explicit.

All 3 tag definition unit tests pass ✅

internal/provider/tag_definition_resource.go
```diff
@@ -357,15 +356,20 @@ func (r *tagDefinitionResource) Create(ctx context.Context, req resource.CreateR
...
+		return nil
+	}
+	elements := make([]types.String, 0, len(list.Elements()))
+	list.ElementsAs(ctx, &elements, false)
+	result := make([]string, len(elements))
+	for i, v := range elements {
 		result[i] = v.ValueString()
 	}
 	return result
 }
```
henryde: f: Using two lists is redundant, you can directly read the values into a string list (and then we should definitely handle the diagnostics).

henryde (changes requested):

internal/provider/tag_definition_resource.go
tob1aswe1ss: Alright. Makes sense.

henryde (approved):

PR #178, opened 2026-05-21
feat: introduce MESHSTACK_SKIP_VERSION_CHECK to skip version check

nroi:
## Summary by CodeRabbit

* **New Features**
  * Added support for bypassing meshStack version compatibility checks via an environment variable override. This allows flexibility when working with different meshStack versions while maintaining existing validation by default.

[Review Change Stack](https://app.coderabbit.ai/change-stack/meshcloud/terraform-provider-meshstack/pull/178?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/meshcloud/terraform-provider-meshstack/client/internal"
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -66,7 +67,11 @@ func New(ctx context.Context, rootUrl *url.URL, userAgent string, auth Authoriza
 	if meshInfo, err := httpClient.GetMeshInfo(ctx); err != nil {
 		return Client{}, fmt.Errorf("failed to retrieve meshStack version information from /mesh/info endpoint: %w", err)
 	} else if meshInfo.Version.Less(MinMeshStackVersion) {
-		return Client{}, fmt.Errorf("unsupported meshStack version: meshStack is running version %s, but this client requires version %s or higher", meshInfo.Version, MinMeshStackVersion)
+		skipVersionCheck := os.Getenv("MESHSTACK_SKIP_VERSION_CHECK") == "true"
```
sttomm, line 70: d: fine for me to fix it like that. But I think we also have to make sure, that we can manually use the latest tf provider also locally ... ok, we have to set this env var then. That's an option. I was just wondering whether we could also determine the host name of the API and skip the version check in case of localhost or meshcloud-dev. But sure, that would be quite ugly specific code her ein the tf provider. So yeah, I guess we will need that env var set then also locally. Seems to be the best option.

sttomm (approved):

client/client.go
nroi: yes, also this can be used as a fallback in case we mess up and accidentally add an incorrect version constraint. Then, we can tell customers to use the env var to disable the version constraint check.

PR #184, opened 2026-05-28
fix: change input/output structure from array to map

nroi:
## Summary by CodeRabbit

* **Refactor**
  * Building block inputs and outputs now use named keys instead of indexed lists for clearer configuration and state mapping.

* **Chore**
  * Minimum required meshStack version raised to 2026.23.0.

* **Documentation**
  * Docs and changelog updated to reflect the new minimum meshStack version and the inputs/outputs shape change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/meshcloud/terraform-provider-meshstack/pull/184?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/buildingblock_v2.go
+++ b/client/buildingblock_v2.go
@@ -36,8 +36,22 @@ type MeshBuildingBlockV2Spec struct {
 	TargetRef                         MeshBuildingBlockV2TargetRef            `json:"targetRef" tfsdk:"target_ref"`
 	DisplayName                       string                                  `json:"displayName" tfsdk:"display_name"`
 
-	Inputs               []MeshBuildingBlockIO     `json:"inputs" tfsdk:"inputs"`
-	ParentBuildingBlocks []MeshBuildingBlockParent `json:"parentBuildingBlocks" tfsdk:"parent_building_blocks"`
+	Inputs               map[string]MeshBuildingBlockV2Input `json:"inputs" tfsdk:"-"`
+	ParentBuildingBlocks []MeshBuildingBlockParent           `json:"parentBuildingBlocks" tfsdk:"parent_building_blocks"`
+}
+
+type MeshBuildingBlockV2Input struct {
+	Value                any     `json:"value"`
+	ValueType            string  `json:"valueType"`
+	IsSensitive          bool    `json:"isSensitive"`
+	AssignmentType       *string `json:"assignmentType"`
+	UpdateableByConsumer bool    `json:"updateableByConsumer"`
+}
+
+type MeshBuildingBlockV2Output struct {
+	Value          any     `json:"value"`
+	ValueType      string  `json:"valueType"`
+	AssignmentType *string `json:"assignmentType"`
 }
 
 type MeshBuildingBlockV2DefinitionVersionRef struct {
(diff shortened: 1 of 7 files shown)


internal/provider/building_block_v2_resource.go
```diff
@@ -340,6 +339,14 @@ func (r *buildingBlockV2Resource) Delete(ctx context.Context, req resource.Delet
 // 	resource.ImportStatePassthroughID(ctx, path.Root("metadata").AtName("uuid"), req, resp)
 // }
 
+func toResourceModelV2Input(key string, io client.MeshBuildingBlockV2Input, diags *diag.Diagnostics) buildingBlockIoModel {
+	return toResourceModel(client.MeshBuildingBlockIO{Key: key, Value: io.Value, ValueType: io.ValueType}, diags)
+}
+
+func toResourceModelV2Output(key string, io client.MeshBuildingBlockV2Output, diags *diag.Diagnostics) buildingBlockOutputModel {
+	return toResourceModel(client.MeshBuildingBlockIO{Key: key, Value: io.Value, ValueType: io.ValueType}, diags).toOutputModel()
```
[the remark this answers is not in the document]
nroi: We don't have this AssignmentType yet in the terraform model. Since the goal of the PR is only to remain compatible with changes in the upstream API, it's okay to leave this as-is for now.

client/buildingblock_v2.go
```diff
@@ -36,8 +36,22 @@ type MeshBuildingBlockV2Spec struct {
 	TargetRef                         MeshBuildingBlockV2TargetRef            `json:"targetRef" tfsdk:"target_ref"`
 	DisplayName                       string                                  `json:"displayName" tfsdk:"display_name"`
 
-	Inputs               []MeshBuildingBlockIO     `json:"inputs" tfsdk:"inputs"`
```
sttomm, line 39: d: How about this MeshBuildingBlockIO now? Do we still need it? Is it used somewhere else? If so, is it good that it is still used there, as we now switch to the map instead of the array? Ah, probably for the V1 version?

internal/provider/building_block_v2_resource.go
```diff
@@ -362,8 +369,8 @@ func setStateFromResponseV2(ctx context.Context, state *tfsdk.State, bb *client.
 	diags.Append(state.SetAttribute(ctx, path.Root("status").AtName("lifecycle"), bb.Status.Lifecycle)...)
 
 	outputs := make(map[string]buildingBlockOutputModel)
-	for _, output := range bb.Status.Outputs {
-		outputs[output.Key] = toResourceModel(output, &diags).toOutputModel()
+	for key, output := range bb.Status.Outputs {
+		outputs[key] = toResourceModelV2Output(key, output, &diags)
 	}
 	if diags.HasError() {
 		return
```
sttomm, line 376: d: Changelog and required meshStack version have to be adapted, right?

internal/provider/building_block_v2_resource.go
sttomm, line 347: d: Why do we use `client.MeshBuildingBlockIO` as a resource model? It is from the client package? And this resource here is the tf resource, is it? And doesn't the tf resource also use maps? Or this actually what will come with the bigger refactoring? For now you just make sure the client layer works with the new meshStack version?

sttomm (approved):

internal/provider/building_block_v2_resource.go
nroi: fixed

client/buildingblock_v2.go
nroi: yes exactly, it's used by the v1 building block.

internal/provider/building_block_v2_resource.go
nroi: yeah we use `client.MeshBuildingBlockIO` basically to just convert it to the legacy format, so we're able to re-use the old function. It's perhaps not the cleanest approach, but we have a huge PR open from Andreas which is going to refactor all of this anyway. So, I'd prefer a small change for now that causes fewer conflicts with Andreas' PR for now.

PR #186, opened 2026-05-29
fix: treat parent_building_blocks as set instead of list to avoid ord…

florianow:
…ering issues

The API may return parent_building_blocks in a different order than sent, causing 'Provider produced inconsistent result after apply' errors. Using SetNestedAttribute makes the comparison order-independent.

Fixes #185

## Summary by CodeRabbit

* **Documentation**
  * Updated building block resource and data source docs to reflect that parent_building_blocks uses Set semantics instead of List.
* **Bug Fixes**
  * parent_building_blocks now behaves as a set (duplicates removed, order not preserved) when stored and validated, aligning behavior with documentation and preventing ordering-related issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/meshcloud/terraform-provider-meshstack/pull/186?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


henryde (approved):
Please don't merge commit

PR #189, opened 2026-05-29
fix: change input/output structure from array to map

nroi:
This is pretty much the same as https://github.com/meshcloud/terraform-provider-meshstack/pull/184

but we had to revert that PR because it was accidentally released to early.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# UNRELEASED
+
+Requires meshStack 2026.23.0 or later (previously 2026.22.0).
+
+BREAKING CHANGES:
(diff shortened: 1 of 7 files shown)


jannymeshi:
📋 Janny has linked this PR to [Feature Shipping Tracker: Extended meshBuildingBlock v3-preview API: Update, Retrigger, and Run Listing](https://github.com/meshcloud/janny/issues/551).

This PR is tracked as the redo of #184 (CU-86c9p4kcd) under DoD item 5 (Terraform Provider Updated). The tracker will be updated when this PR merges.

sttomm (approved):

PR #190, opened 2026-06-02
fix: align BB v2 with embedded-secret inputs and add purge_on_delete

grubmeshi:
## Summary

Aligns the `meshstack_building_block_v2` client with the upstream meshStack API's **embedded-secret** representation for sensitive inputs, and adds a new **`purge_on_delete`** attribute. Requires meshStack **2026.23.0+**.

### 1. SecretEmbedded support (read-side fix)
The v2-preview API now returns sensitive inputs as embedded secrets (`{"hash":"sha256:…"}` with `isSensitive:true`). Previously the value deserialized into a map and was **silently dropped** from `combined_inputs`. 

- `MeshBuildingBlockV2Input.Value` is now `types.SecretOrAny`, with an `UnmarshalJSON` that routes sensitive values into the secret branch and plain values into the scalar branch (mirroring `MeshBuildingBlockDefinitionInput`).
- `toResourceModelV2Input` surfaces the secret hash into state instead of dropping it.
- No Terraform schema change; setting sensitive *user* inputs remains out of scope.

### 2. `purge_on_delete` attribute
New optional `purge_on_delete` (default `false`). When `true`, deletion calls the API's `/purge` endpoint, removing the Building Block without running its configured deletion run — useful when a Building Block is stuck in a non-final state. Requires `ADM_BUILDINGBLOCK_DELETE`. Deletion now always polls for completion so dependent resources can be cleaned up safely.

## Tests
- New live acceptance subtest `TestAccBuildingBlockV2/03_sensitive_input` exercises the embedded-secret read path end-to-end using a STATIC sensitive input (no provider write support needed), with `purge_on_delete = true` so the stuck PENDING block is removed cleanly.
- Verified against a local meshStack: BB v2 + runner acceptance tests pass.

## Incidental test-fixture fixes
Two pre-existing broken acceptance tests (surfaced while running the full suite) are fixed here:
- Manual BB v2 BBD fixtures now declare an `environment` output (manual BBDs mirror inputs→outputs server-side).
- Building block runner acceptance tests now inject a valid embedded RSA public key instead of the truncated doc placeholder.

## Release Notes v0.21.0

* **New Features**
  * Added optional `purge_on_delete` parameter to meshstack_building_block_v2 resource for immediate deletion capability

* **Changes**
  * Enhanced sensitive input handling in building block v2
  * Updated deletion behavior to consistently wait for completion confirmation


jannymeshi:
📋 **PR linked to tracker** — [#551 Extended meshBuildingBlock v3-preview API](https://github.com/meshcloud/janny/issues/551)

This PR (embedded-secret read fix + `purge_on_delete`) is tracked under **DoD item 5 (Terraform Provider Updated)** in tracker #551. It will be checked off once merged.

nroi (approved):

PR #202, opened 2026-06-15
fix: disable Go's test result cache for acceptance tests

nroi:


Taskfile.yml
```diff
@@ -8,8 +8,9 @@ tasks:
     env:
       TF_ACC: 1
     cmds:
-      - go test ./... -v {{.CLI_ARGS}} -timeout 120m
-    
+      # -count=1 disables Go's test result cache — skipping is handled at the Gradle level based on source inputs
```
henryde: t: A bit confusing since there is no gradle in this repo.

henryde (approved):

Taskfile.yml
nroi: valid point, I've changed it.

PR #207, opened 2026-06-23
ci: dockerized acceptance test workflow

grubmeshi:
## What

Adds a dockerized **acceptance-test job** to the `Tests` workflow (`.github/workflows/test.yml`) that brings up the **entire meshStack backend from pre-built images** as GitHub Actions service containers on the self-hosted `mesh-runners`, then runs the provider acceptance suite (`go tool gotestsum … ./internal/provider/…`) against it and merges its coverage with the unit job's.

Stack (all pulled, none built in this repo):
- infra: `mariadb-ci`, `ravendb-ci`, `keycloak-ci:26`, `rabbitmq`
- services: `meshfed` (8089), `replicator`, `block-coordinator-api`
- building-block fan-out: `multiplexing-block-runner-ci` (mux) + `tf-block-runner` + `manual-block-runner`

## Design notes

- Internal registry path comes from the **`MESHFED_REGISTRY` repo configuration variable** (not hardcoded in this public repo). `env` context isn't allowed in `services.image`, so a `vars.*` configuration variable is used.
- meshfed-api runs on **8089** so the tf-runner's fixed `:8080` health server doesn't collide (matches meshfed-release CI).
- JVM services use `--restart on-failure` to absorb the mariadb startup race; a readiness step gates the tests.
- Runners authenticate with the seeded **managed-runners** API key (runner-scoped rights), not the provider admin key.
- Coverage: the unit job's coverage data is merged with the acceptance job's into the single PR coverage comment.
- The canonical local equivalent is `meshfed-release/docker-compose.acc.yml` (private repo); this workflow is the native-`services:` translation.
- Runs on non-fork PRs and `workflow_dispatch` only (forks lack self-hosted runners / the variable).

## Advisory (non-blocking) gating

The acceptance job runs against the **last merged** meshfed-release backend images, so a provider change that needs a companion backend change can't go green here until that backend merges. To avoid a cross-repo merge deadlock, this job is **advisory**:

- `continue-on-error` keeps the `Go Acceptance Test` check green regardless of result (it is not a required check).
- When the latest run fails, a PR comment is posted with a link to the run log; it is auto-deleted once a later run goes green again (idempotent via an HTML marker).

The authoritative cross-repo gate is meshfed-release's source-built `terraform-provider-acceptance` job, which pairs the same-named provider branch and builds the backend from that PR's source.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,11 @@ on:
 permissions:
   contents: read
 
+# Cancel superseded runs on the same ref (the acceptance job is expensive).
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Go Build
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -90,44 +95,317 @@ jobs:
...
+          echo "mux routing:"; curl -s http://localhost:8309/status || true
+
+      - name: Run acceptance tests with coverage
+        id: acc
+        # Advisory: a failed acceptance run must never block the PR. The job runs against the last
+        # *merged* meshfed-release backend images, so a change needing a companion backend PR cannot
+        # go green here until that backend merges — gating on it would deadlock cross-repo changes.
+        # meshfed-release's source-built `terraform-provider-acceptance` job is the real gate; here we
+        # only surface failures via an advisory comment (see the last step).
+        continue-on-error: true
```
nroi, line 281: i: make sure to follow up on this PR and check that this step then ends up being green unless there are any errors. Otherwise, this will turn into yet another noisy signal where everyone just says "yeah this fails for whatever reason, but we ignore it" 😬

nroi (approved):
lgtm

.github/workflows/test.yml
grubmeshi: well, it's intentionally done this way and explicitly reported in PRs as sometimes PR also need backend changes, so the tf acc tests in meshfed-release are the authority to tell if that all works here.

PR #208, opened 2026-06-24
feat: meshstack_building_block resource + meshstack_building_blocks data source (building block v3)

grubmeshi:
Adds the `meshstack_building_block` resource (v3, preview API) for managing Building Blocks via the meshObject API.

## What it delivers

- Generic, typed inputs via `jsonencode(...)`; sensitive inputs via a `sensitive` block with `secret_version`-controlled rotation.
- In-place version upgrade by changing `building_block_definition_version_ref.uuid`.
- `moved` support from `meshstack_buildingblock` (v1) and `meshstack_building_block_v2`.
- Status surfaced — `status.latest_run_uuid` (latest modifying run) and the new computed `status.latest_dry_run_uuid`; runs awaited with run-log diagnostics on failure, with a configurable `timeouts` block (create/update/delete).

## Behavior contract

- **Replacement:** only `target_ref.*` forces replacement (you cannot move a Building Block to another tenant/workspace). `display_name` and `inputs` update in place. `parent_building_blocks` change only as part of a version upgrade — changing them on their own forces replacement.
- **Reruns:** a run is triggered on a version change, an input change, a sensitive-input rotation (`secret_version` bump), a `content_hash` change, or a parent change carried by a version upgrade; a rename or a no-op does not. An update that supplies a previously-missing input (unblocking a block parked in `WAITING_FOR_OPERATOR_INPUT`) awaits the resulting run instead of returning early.
- **Partial / shared input ownership:** a configuration manages only the inputs it declares in `spec.inputs`; an input set by another party but omitted is preserved server-side and surfaced read-only in `all_inputs`, not as drift. This lets a platform operator manage operator inputs while the consumer owns user inputs, each declaring only its own.
- **Upgrades:** only to the latest released version of the same definition.
- **Sensitive inputs:** use the `sensitive` block iff the definition declares the input sensitive; a `moved` migration warns to re-supply the secret (it cannot travel through state).
- A version change on a non-completed block, and operator inputs set by a non-operator key, surface clear errors instead of raw conflicts or silent drift.

## Data source: `meshstack_building_blocks`

Adds a read-only, filterable companion list data source (backed by the same v2-preview list endpoint). Each returned block mirrors the resource schema (`metadata`/`spec`/`status`/`all_inputs`); sensitive inputs are surfaced as a hash only.

- Filters: `workspace_identifier`, `project_identifier`, `platform_identifier`, `name`, `definition_uuid`, `version_uuid`, `version_number` (lenient `v1`/`1`), `tenant_uuid`, `target_kind`, `status`, `lifecycle_states` (repeated param; empty ⇒ all), and the platform-operator scope selectors `managed_by_definition_uuid` / `managed_by_workspace_identifier` (`MANAGED_BUILDINGBLOCK_LIST`).
- `version_number` filtering requires the paired meshfed change (meshcloud/meshfed-release#10063).

## What it enables — app team & operator use cases

A local end-to-end demo (operator → app team → cross-workspace upgrade) exercises the new building blocks this resource unlocks:

**App team (workspace-scoped consumer)**
- Self-serves a building block from a released definition discovered **by name** (`meshstack_building_blocks` / `meshstack_building_block_definitions` data sources), pinning to a chosen version.
- Owns **only its own inputs**: it declares user inputs; operator inputs and any inputs set by others are preserved server-side and surfaced read-only in `all_inputs`, never as drift.
- **Cannot change the definition version** (operator/admin-only — a self-attempted upgrade is rejected 403) and **cannot see run logs** when the definition has `run_transparency = false` (`status.latest_run_uuid` comes back null).
- **Reconciles an external upgrade**: when the operator bumps the version cross-workspace, the app team sees it as drift and accepts it by repinning — the block is updated in place, not destroyed/recreated.

**Platform operator (definition owner)**
- Authors and **tests a definition in its own workspace** with a real **sensitive input decrypted end-to-end** before exposing it.
- Editing a draft's module re-runs the existing test block **in place** via `content_hash` (no replace, no teardown), surfacing the failing run's logs as the definition owner even when transparency is off.
- **Adopts and upgrades app-team blocks cross-workspace** with a `MANAGED_…` key: `import` blocks keyed off the data source pull each block made from its definition into operator state, then a pure version bump upgrades them — **without owning, and never deleting** them (teardown uses `lifecycle { destroy = false }` to forget, not delete).
- A **defaulted operator input** lands automatically on upgrade (empty `inputs = {}` → backend fills the default), and the app team's user inputs (including the sensitive one) are left untouched.

## Also includes

- `meshstack_landingzone`: `spec.platform_ref` and `spec.mandatory_building_block_refs`/`spec.recommended_building_block_refs` now accept a referenced resource's computed `ref` directly (e.g. `platform_ref = meshstack_platform.example.ref`); `kind` is optional rather than read-only, so assigning a full ref object no longer fails.
- meshObject UUID references now validate at plan time: a ref object that is provided but missing its `uuid` is rejected up front (instead of only failing against the backend), while an assigned computed `.ref` — whose `uuid` is unknown until apply — keeps working. Applies to the landingzone building-block refs and the building block definition's `dependencies` / `integration_ref`.

---

Backend counterpart: meshcloud/meshfed-release#10063.


client/building_block_v2.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// IsWaitingForInput reports whether the building block run is paused awaiting
+// human or dependency input. Such a run will not progress on its own, so polling
+// callers treat it as a terminal (but non-fatal) state and surface a warning.
+func (bb *MeshBuildingBlockV2) IsWaitingForInput() bool {
+	return bb.Status.Status == BuildingBlockStatusWaitingForOperatorInput ||
+		bb.Status.Status == BuildingBlockStatusWaitingForUserInput ||
+		bb.Status.Status == BuildingBlockStatusWaitingForDependentInput
+}
```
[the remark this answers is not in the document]
grubmeshi: this is bogus, the backend never returns nil status.

internal/types/secret/secret.go
```diff
@@ -45,12 +55,9 @@ func WithConverterSupport(ctx context.Context, config, plan, state generic.Attri
...
 func WithDatasourceConverter() generic.ConverterOption {
-	type datasourceSecret struct {
-		Hash string `tfsdk:"secret_hash"`
-	}
-	return generic.WithValueFromConverterFor[clientTypes.Secret](generic.ValueFromConverterForTypedNilHandler[datasourceSecret](),
+	return generic.WithValueFromConverterFor[clientTypes.Secret](generic.ValueFromConverterForTypedNilHandler[HashOnly](),
 		func(attributePath path.Path, in clientTypes.Secret) (tftypes.Value, error) {
-			return generic.ValueFrom(datasourceSecret{Hash: *in.Hash})
+			return generic.ValueFrom(HashOnly{Hash: *in.Hash})
 		})
```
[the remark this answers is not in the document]
grubmeshi: also false positive / bogus

internal/provider/acctest/testconfig/build_building_block.go
```diff
@@ -0,0 +1,73 @@
...
+		Descend("spec", "target_ref")(SetAddr(workspaceAddr, "ref")),
+	).Join(workspaceConfig, buildingBlockDefinitionConfig), buildingBlockAddr, buildingBlockDefinitionAddr, workspaceAddr
+}
+
+// BBTenant builds a workspace (+project/platform/landing-zone/tenant) and a v3 building block
+// targeting that tenant. The building block definition uses the terraform implementation; callers
+// pass terraformRepoUrl (a file:// URL to the committed bare repo) so the real tf-block-runner can
+// clone and run OpenTofu offline in acceptance mode. In mock mode the URL is unused. workspaceAddr
+// is the underlying Workspace(t) address, returned so callers can attach further workspace-scoped
+// resources without rebuilding the workspace.
```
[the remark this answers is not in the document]
grubmeshi: Fixed in 2b3743d — the comment now describes the repo as served over a loopback git smart-HTTP URL by the test's git-http-backend.

internal/provider/acctest/testconfig/build_building_block.go
```diff
@@ -0,0 +1,73 @@
...
+	)
+
+	var buildingBlockDefinitionAddr Traversal
+	buildingBlockDefinitionConfig := Resource{Name: "building_block", Suffix: "_02_tenant"}.TestSupportConfig(t, "").WithFirstBlock(
+		ExtractAddress(&buildingBlockDefinitionAddr),
+		OwnedByWorkspace(workspaceAddr),
+		Descend("spec", "supported_platforms")(SetRawExpr("[{name = %s}]", platformTypeAddr.Join("metadata", "name"))),
+		// Point the terraform implementation at the committed local bare repo (file://) so the real
+		// tf-block-runner clones and runs OpenTofu offline. The static example URL in the .tf is a
+		// docs placeholder; in mock mode this value is unused.
```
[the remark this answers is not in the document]
grubmeshi: Fixed in 2b3743d — comment updated to reference the loopback git smart-HTTP serving instead of file://.

internal/provider/git_http_server_test.go
```diff
@@ -0,0 +1,116 @@
...
+		return "", err
+	}
+	backend, err := gitHTTPBackendPath()
+	if err != nil {
+		return "", err
+	}
+	// Bind 0.0.0.0 so peers in the pod/host can connect; advertise 127.0.0.1 (loopback is shared
+	// across containers in the CI pod and is the host loopback for a local `go run` runner).
+	ln, err := net.Listen("tcp", "0.0.0.0:0")
+	if err != nil {
```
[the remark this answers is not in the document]
grubmeshi: Keeping the 0.0.0.0 bind intentionally. In CI the tf-block-runner runs in a separate container and reaches this fixture over the pod network; binding only to 127.0.0.1 risks making it unreachable cross-container, which the cross-repo acceptance wiring depends on. It is a throwaway server on an ephemeral port for the test's lifetime, and the advertised clone URL stays 127.0.0.1, so exposure is limited. The rationale is documented inline at the bind site.

internal/provider/landingzone_resource.go
```diff
@@ -68,7 +68,7 @@ func (r *landingZoneResource) Schema(_ context.Context, _ resource.SchemaRequest
 	}
 
 	buildingBlockRefs := schema.NestedAttributeObject{
-		Attributes: meshBuildingBlockDefinitionRefAttribute(false),
+		Attributes: meshUuidRefAttribute(client.MeshObjectKind.BuildingBlockDefinition),
 	}
```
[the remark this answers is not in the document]
grubmeshi: Good catch. We deliberately keep `uuid` Optional+Computed in the shared `meshUuidRefAttribute` so a whole computed `.ref` object (whose `uuid` is unknown at plan time) can be assigned — making `uuid` Required regresses that (it rejects the unknown value: `mandatory_building_block_refs = [bbd.ref]` fails to plan). Instead, fixed in 999a4f5 with a plan-time `RequiredAttributesValidator` that rejects a ref object that is *known but missing* `uuid`, while still skipping null/unknown objects and unknown attribute values (so an assigned `.ref` keeps working). Wired into landingzone mandatory/recommended_building_block_refs and the BBD dependencies/integration_ref; runner refs keep optional uuid (backend fills the shared-runner default).

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "meshstack_building_block Resource - terraform-provider-meshstack"
+subcategory: ""
+description: |-
+  Manage a workspace or tenant building block using the meshBuildingBlock v2-preview API transport.
```
grubmeshi: f: schema docs should not mention API explicitly, as TF provider aims to hide api versioning concerning resource design. see also other places with similar issue in docs and fix this up.
```suggestion
  Manage a workspace or tenant building block created from a building block definition.
```

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Optional
+
+- `purge_on_delete` (Boolean) When true, deletes via the `DELETE /{uuid}/purge` sub-path, which requires admin authority (`ADM_BUILDINGBLOCK_DELETE`). This is a last resort option for stuck deletions.
+- `timeouts` (Attributes) (see [below for nested schema](#nestedatt--timeouts))
+- `wait_for_completion` (Boolean) Whether to wait for the building block to reach a terminal state (SUCCEEDED or FAILED) before completing create/update operations. The provider emits actionable warnings if the run is blocked in `WAITING_FOR_OPERATOR_INPUT`. Deletion always waits for the block to be fully removed (lifecycle DELETED, or gone after a purge) regardless of this flag, so dependent resources such as the building block definition can be deleted safely afterward.
+
+### Read-Only
+
+- `all_inputs` (Attributes Map) Computed read-only view of all inputs. Shows the plain value for non-sensitive inputs or the hash for sensitive inputs. Set values for them in `spec.inputs`. (see [below for nested schema](#nestedatt--all_inputs))
```
grubmeshi: f: explain here that "all" means platform operator, user inputs and static inputs and other sources. 

note the relation to the other attribute `inputs` which only defines the inputs managed by this resource, and that operators manage `inputs` only for platform operators  / (inderitcly static ones via bbd as well) without managing the whole resource usually (typical cross workspace app teams are managing the BB lifecycle, with the exception that operators create BB from draft BBds for testing). This situation could be explained in a bit more detail as part of the resource description, but don't become too chatty / stay succinct.

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+- `target_ref` (Attributes) References the building block target. For `meshTenant` targets, `uuid` is required and `name` must be omitted. For `meshWorkspace` targets, `name` is required and `uuid` must be omitted. (see [below for nested schema](#nestedatt--spec--target_ref))
+
+Optional:
+
+- `parent_building_blocks` (Attributes Set) List of parent building blocks. Parent building blocks can only change as part of a version upgrade; changing them on their own forces the building block to be replaced (destroyed and recreated). (see [below for nested schema](#nestedatt--spec--parent_building_blocks))
```
grubmeshi: d: this description could include that parent relations should align with inputs whose values are taken from output of the BB of that parent. this is also a general concept of BBs that they can build a dependency hierarchy. think about adding this to the BB description as well, or look up and include reference docs such as https://docs.meshcloud.io/concepts/building-block/ focus on the more technical aspects when it comes to managed BBs as code with TF here.

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+<a id="nestedatt--spec--building_block_definition_version_ref"></a>
+### Nested Schema for `spec.building_block_definition_version_ref`
+
+Required:
+
+- `uuid` (String) UUID of the building block definition version. Must reference the latest released version of the definition when upgrading.
+
+Optional:
+
+- `content_hash` (String) Content hash of BBD version. If changed, triggers a new run of the building block. After `terraform import`, the first apply triggers a run if `content_hash` is set in config (because import leaves it null in state). To avoid an unexpected run after import, either omit `content_hash` from config until after the first post-import apply, or set it only after the import completes.
```
grubmeshi: please validate if that statement is actually true from code and ensure that in code, it's mentioned that content_hash description mention this (cross link). also, focus on the fact that content_hash only exists to detect changes in draft BBDs to concentiently re-run a BB. it can also be "misused" to trigger a rerun manually, but that should be used with care and needs run_transparency AFAICT to trigger with mere app team permission (BB_SAVE unprefixed for single workspace, no ADM / MANAGED)

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+
+
+
+<a id="nestedatt--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours).
+- `delete` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours). Setting a timeout for a Delete operation is only applicable if changes are saved into state before the destroy operation occurs.
```
grubmeshi: f: i don't understand the part bout to what that timeout applies. why not simply make that timeout for delete operation itself and if it times out, the resource gets removed from state but it errors with a clear warning that managed resources might be left over?

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+- `content_hash` (String) Content hash of BBD version. If changed, triggers a new run of the building block. After `terraform import`, the first apply triggers a run if `content_hash` is set in config (because import leaves it null in state). To avoid an unexpected run after import, either omit `content_hash` from config until after the first post-import apply, or set it only after the import completes.
+
+
+<a id="nestedatt--spec--inputs"></a>
+### Nested Schema for `spec.inputs`
+
+Optional:
+
+- `sensitive` (Attributes) Sensitive input value. Use `secret_value` for create/rotate and `secret_version` to control updates. (see [below for nested schema](#nestedatt--spec--inputs--sensitive))
+- `value` (String) Non-sensitive input value. Use `jsonencode(...)` for non-string values.
```
grubmeshi: f: this is not quite right, as even string values must be `jsonencode`, see BBD description and compare. fix BBD description if it's also not clear there. also, the general description of the BB resource should mention that it's similar to how the BBD inputs are defined (but BBD is a bit more complex concerning default values etc.)

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+- `display_name` (String) Display name for the building block as shown in meshPanel. Changing it is applied in place (a rename) and does not trigger a building block run.
+- `inputs` (Attributes Map) Input values for this building block. Normal app-team users only provide `USER_INPUT` inputs here. `PLATFORM_OPERATOR_MANUAL_INPUT` only makes sense when a platform operator creates a BB from a BBD in the same workspace (e.g. to test a draft BBD before release), using an API key with sufficient permissions; providing an operator input with a non-operator/non-owner API key is rejected with a clear error. Set either `value` using `jsonencode(...)` or `sensitive = { secret_value = ... }`. Use the `sensitive` block if and only if the BBD declares the input as sensitive — a mismatch in either direction is rejected by the backend. (see [below for nested schema](#nestedatt--spec--inputs))
```
grubmeshi: use paragraphs with `<br/>` (check how other longer descriptions are made). stay succinct regarding BBD sensitive requirement (it's clear that its a bug and consider starting with this instead of the appteam/operator distinction)

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Required
+
+- `spec` (Attributes) Building block specification. (see [below for nested schema](#nestedatt--spec))
+
+### Optional
+
+- `purge_on_delete` (Boolean) When true, deletes via the `DELETE /{uuid}/purge` sub-path, which requires admin authority (`ADM_BUILDINGBLOCK_DELETE`). This is a last resort option for stuck deletions.
+- `timeouts` (Attributes) (see [below for nested schema](#nestedatt--timeouts))
+- `wait_for_completion` (Boolean) Whether to wait for the building block to reach a terminal state (SUCCEEDED or FAILED) before completing create/update operations. The provider emits actionable warnings if the run is blocked in `WAITING_FOR_OPERATOR_INPUT`. Deletion always waits for the block to be fully removed (lifecycle DELETED, or gone after a purge) regardless of this flag, so dependent resources such as the building block definition can be deleted safely afterward.
```
grubmeshi: mention timeouts here but also see my comment regarding deletion timeout there.

docs/resources/building_block.md
```diff
@@ -0,0 +1,280 @@
...
+- `owned_by_workspace` (String) The workspace containing this building block.
+- `uuid` (String) UUID which uniquely identifies the building block.
+
+
+<a id="nestedatt--status"></a>
+### Nested Schema for `status`
+
+Read-Only:
+
+- `force_purge` (Boolean) Indicates whether an operator has requested purging of this Building Block.
```
grubmeshi: investigate again if exposing this in the tf repo makes sense (find out in backend what this flag actually indicate validated by understanding the code). also relate this to the purge_on_delete flag, which should at least be mentioned if we keep exposing this flag for now. main quesion is is there a difference between force purge and the purge on delete functionality? It might be that purge_on_delete can only be done by platform operator? also then renaming this to simply `purged` is better as a status field (also align backend then), this is similar to the lifecycle status which becomes DELETED even if it was purged (so resources went unmanaged, no destroy run triggered for BB)

docs/resources/building_block.md
grubmeshi: Validated against code: `content_hash` is Terraform-only and a content_hash-only change forces a run via the trigger-run endpoint. Reframed the description (detect draft-BBD content changes + manual-rerun-with-care) and added the run-transparency note — a plain workspace key (`BUILDINGBLOCK_SAVE`) needs the BBD's run transparency enabled; admins and the definition's operator are exempt.

docs/resources/building_block.md
grubmeshi: Replaced the generic framework blurb with what the timeout actually bounds: the deprovision wait (async destroy run, or removal after a purge); on timeout the apply errors and keeps the resource in state for retry. Your 'remove from state + warn on timeout' idea is a behavior change — tracked as a proposal; I'd keep the safer retain-and-error default (it protects dependent deletes), so leaving it as a possible opt-in later.

docs/resources/building_block.md
grubmeshi: Improved the description and related it to `purge_on_delete`. Verified in backend: the force-purge endpoint currently requires `ADM_BUILDINGBLOCK_DELETE` only. Your `purged` rename + allowing `MANAGED_BUILDINGBLOCK_DELETE` (operator owning the BBD) is tracked as a meshfed-release proposal (separate commit + test + release note).

internal/clientmock/mock_building_block_v2.go
```diff
@@ -0,0 +1,276 @@
...
+	newRunUuid := uuid.NewString()
+	if cp.Status == nil {
+		cp.Status = &client.MeshBuildingBlockV2Status{
+			Lifecycle: client.MeshBuildingBlockV2Lifecycle{
+				State: client.BuildingBlockLifecycleStateActive,
+			},
+		}
+	}
+	cp.Status.Status = client.BuildingBlockStatusSucceeded
+	cp.Status.LatestRunUuid = &newRunUuid
```
nroi: d: I get the following hint in intellij:

> Syntax update in Go 1.26: replace pointer to local variable 'newRunUuid' with new()

internal/clientmock/mock_buildingblock.go
```diff
@@ -2,45 +2,217 @@ package clientmock
...
+			Uuid:              id,
+			DefinitionUuid:    definitionUuid,
+			DefinitionVersion: definitionVersion,
+			TenantIdentifier:  tenantIdentifier,
+			CreatedOn:         mockBuildingBlockCreatedOn,
+		},
+		Spec: client.MeshBuildingBlockSpec{
+			DisplayName:          v2.Spec.DisplayName,
+			Inputs:               inputs,
+			ParentBuildingBlocks: []client.MeshBuildingBlockParent(v2.Spec.ParentBuildingBlocks),
```
nroi: d: "Redundant type conversion" according to intellij:

```suggestion
			ParentBuildingBlocks: v2.Spec.ParentBuildingBlocks,
```

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1240 @@
...
+						Attributes: map[string]schema.Attribute{
+							"uuid": schema.StringAttribute{
+								MarkdownDescription: "UUID of the building block definition version. Must reference the latest released version of the definition when upgrading.",
+								Required:            true,
+							},
+							"content_hash": schema.StringAttribute{
+								MarkdownDescription: "Content hash of the building block definition version. " +
+									"Its purpose is to detect content changes of a **draft** BBD (whose version `uuid` stays the same) " +
+									"and conveniently re-run the building block when it changes.<br>" +
+									"It is Terraform-only and never sent to the backend, so changing it can also be used to force a " +
```
nroi: d: shouldn't we use the term OpenTofu here? iirc it's something we have to be careful with after Terraform went closed source. Also it might be more consistent to use the term OpenTofu here (although I did not check in detail tbh), but in the panel for example, when creating a new BBD, we use the term OpenTofu.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1240 @@
...
+									"manual re-run — use with care; with a plain workspace key (`BUILDINGBLOCK_SAVE`) this requires the " +
+									"definition to have run transparency enabled (admins and the definition's platform operator are exempt).<br>" +
+									"After `terraform import` it is left null in state, so the first apply triggers a run if `content_hash` " +
```
nroi: d: similar to the previous comment, maybe replace by tofu?

```suggestion
									"After `tofu import` it is left null in state, so the first apply triggers a run if `content_hash` " +

```

also check for similar occurrences of the term terraform / Terraform.

internal/provider/building_block_resource.go
```diff
@@ -0,0 +1,1240 @@
...
+									// further enforces which one is required for the given kind.
+									stringvalidator.ExactlyOneOf(
+										path.MatchRelative().AtParent().AtName("uuid"),
+										path.MatchRelative().AtParent().AtName("name"),
+									),
+								},
+								PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+							},
+							"name": schema.StringAttribute{
+								MarkdownDescription: "Identifier (name) of the target workspace. Required when `kind = \"meshWorkspace\"`, must be omitted for `kind = \"meshTenant\"`.",
```
nroi: d: I always found this identifier/name business very confusing and unintuitive in our meshObject API. `name` raises the expectation that it's something human-readable *and* mutable that may contain spaces and such, which is not the case with the `name` in our meshObject API. The fact that the workspace has both a name and a displayName (which in the panel is referred to as identifier and name, respectively) adds to the confusion.

I'd just remove the `(name)` here.

```suggestion
								MarkdownDescription: "Identifier of the target workspace. Required when `kind = \"meshWorkspace\"`, must be omitted for `kind = \"meshTenant\"`.",

```

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+# v0.23.0
```
nroi, line 1: d: in addition to a list of new features and fixes, it might be helpful to also have an introductory sentence to explain the big picture here regarding this massive change. And also to provide some guidance. Because we have now:
* `meshstack_buildingblock`
* `meshstack_building_block_v2`
* `meshstack_building_block`

we should explain why we have introduced yet another resource and which one should be used.

nroi: also, the bullet points are a bit weird, we have one bullet point to announce the new `meshstack_building_block` resource, and then a whole bunch of bullet points describing features and fixes for this resource. maybe don't stick to this existing FEATURES / FIXES format and just describe what this new resource can do that wasn't possible before. Also it's odd to have FIXES for a new resource that was just recently created.

nroi (approved):

internal/clientmock/mock_building_block_v2.go
grubmeshi: Yeah, there's actually a modern go skill here and I wonder why that didn't trigger. Everything needs to be linted :(

internal/provider/building_block_resource.go
grubmeshi: yes, i think we should shorten down the description here and focus on the purpose of `content_hash` property

grubmeshi: Rather than switch to `tofu`/`OpenTofu`, I dropped the vendor term here: `content_hash` is now described as "provider-only … never sent to the backend", which is also more accurate and sidesteps the trademark concern without tying the doc to one CLI. If you'd rather standardize on `tofu`/`OpenTofu` to match the panel, I'm happy to switch — just say the word.

internal/provider/building_block_resource.go
grubmeshi: Same approach here: "After import …" (dropped the CLI qualifier). I also swept the remaining `Terraform` mentions in `building_block_resource.go` (the two code comments) to neutral wording. I left the established `Terraform` phrasing in the other, pre-existing resources (platform/landingzone/…) and the BBD release-approval warning untouched — they predate this PR and use `Terraform` consistently, so a broad rename felt out of scope here. Can do a separate sweep for those if you'd like.

CHANGELOG.md
grubmeshi: Kept this as a single concise `FEATURES` entry rather than a separate intro paragraph, but it carries the "big picture" guidance you asked for: it names `meshstack_building_block` as "the recommended way to manage building blocks", states the *why* (updates in place instead of forcing a destroy-and-recreate), and gives the migration path from both older resources (`meshstack_buildingblock` and `meshstack_building_block_v2`) via a `moved` block. So "which one to use and why" is covered without a chatty preamble. Does that address the concern, or would you prefer an explicit intro sentence on top?

PR #210, opened 2026-06-25
ci: run acceptance on main (non-advisory) + fix Entra ID acceptance test ownership

grubmeshi:
## Why

Two related gaps surfaced by the failing `meshfed-release` `terraform-provider-acceptance` run on `develop`, which exercises this provider's new Entra ID acceptance test:

1. **Acceptance tests were skipped on `main`.** The acceptance job only ran on non-fork PRs and manual dispatch, so a provider/backend regression was invisible in this repo and could ship in a release tag before anyone noticed.
2. **The new Entra ID acceptance test was wrong.** It owned the integration by a freshly created test workspace, but meshfed only permits Entra ID integrations on the **admin (partner) workspace** — so create failed.

## What

**CI (`test.yml`)**
- Run the acceptance job on pushes to `main` too (`github.event_name == 'push'`).
- Make the run advisory (`continue-on-error`) **only on PRs**. On `main`/dispatch a failure reds the job, surfacing regressions before a release tag — the same failure `meshfed-release`'s source-built acceptance job hits. PRs keep the advisory comment behavior (they can legitimately need a companion backend PR before going green).
- Select acceptance tests by the `TestAcc` **naming convention** across all packages (`-run TestAcc ./...`) instead of restricting to `./internal/provider`, so future acceptance tests in any package are picked up automatically.

**Entra ID acceptance test (`04_entra_id`)**
- Own the integration by the dev-dump-seeded `demo-partner` admin workspace via a new `AdminWorkspaceIdentifier` constant (in `provider_test.go`) and a new `testconfig.IntegrationForWorkspace` builder, instead of creating a test workspace.
- Drop the `display_name` update step: meshfed's `EntraIdIntegrationHandler.update` ignores `model.name`, so renaming is silently dropped by the backend and surfaces as `inconsistent result after apply`. Documented inline; the step can be restored once the backend persists the name.

## Backend companion PR

The rename bug is fixed in meshfed-release: https://github.com/meshcloud/meshfed-release — branch `fix/entra-id-integration-rename`. No provider behavior depends on it; this PR is mergeable independently.

## Testing

- `task lint` — clean
- mock-mode unit tests — pass
- acceptance tests against the local dev-dump stack — `TestAccIntegration*` pass (create + import for Entra ID)

No CHANGELOG entry: test/CI-only changes, no user-facing provider behavior change (the Entra ID feature itself is already in v0.22.1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #211, opened 2026-06-25
test: verify Entra ID integration can be renamed

grubmeshi:
## What

Follow-up to #210. Restores the `display_name` update step in the `04_entra_id` acceptance subtest so it verifies an Entra ID integration can be **renamed** (create → rename → import).

## Why now

#210 dropped that step because meshfed's `EntraIdIntegrationHandler.update` silently ignored `model.name` (rename was a no-op, surfacing as `inconsistent result after apply`). The backend fix is meshcloud/meshfed-release#10225, which persists the name on update.

## Cross-repo pairing

This branch is named **`feature/entra-id-integration-rename`** to match the meshfed-release PR branch. meshfed-release's `terraform-provider-acceptance` job resolves the provider branch from `github.head_ref` and falls back to `main` — so with matching names it pairs the two and validates the rename against the **fixed** backend.

> ⚠️ This provider PR's own acceptance job runs against the last *merged* meshfed-release backend images, which do **not** yet contain the fix, so the rename step will fail here (advisory-only on PRs) until #10225 merges and ships in the images. The authoritative signal is the paired run in meshfed-release#10225.

Mock-mode unit tests and lint pass.

### DIFF
--- a/internal/provider/integration_resource_test.go
+++ b/internal/provider/integration_resource_test.go
@@ -203,12 +203,18 @@ func TestAccIntegrationResource(t *testing.T) {
 						xknownvalue.Ref(resourceAddress, "meshIntegration", &resourceUuid),
 					},
 				},
-				// NOTE: no display_name update step here (unlike the other integration types).
-				// meshfed's EntraIdIntegrationHandler.update ignores model.name — it updates only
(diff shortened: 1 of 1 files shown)


j0g3sc (approved):
can only be released after meshfed is released before

PR #218, opened 2026-07-03
feat: allow to manage display_order value for building block definiti…

j0g3sc:
…on I/O without affecting the calculated hash for change detection

CU-86cabn76y

---
### Summary
Added a manageable **display_order** field to building block definition version inputs and outputs.

**Implementation**:

**Client**
DisplayOrder int64 with json:"displayOrder,omitempty" on both MeshBuildingBlockDefinitionInput and MeshBuildingBlockDefinitionOutput.

**Schema**
 display_order Int64Attribute on inputs and outputs, Optional + Computed + Default(0) — the Default is what keeps re-plans clean given the backend returns 0 when unset.

**Content hash** 
stripped to 0 before hashing, so reordering never changes content_hash or forces a new version

**Released immutability** 
 displayOrdersDiffer guard in Update — needed because display_order is outside the hash, so this restores the "released versions are immutable" behavior with a clear error.

**Mock**
mirrors input→output display_order for manual BBDs.

---

:warning: depends on meshfed-api released with https://github.com/meshcloud/meshfed-release/pull/10263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # v0.23.1
 
 FEATURES:
-- `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to `0` when omitted and is intentionally excluded from a version's `content_hash`, so reordering does not create a version change.
+- `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to `0` when omitted.
 
 FIXES:
 - `meshstack_building_block`: Prepare for the upcoming `WAITING_FOR_APPROVAL` run status (an approval gate coming soon to meshStack — not available yet). Once meshStack starts returning it, an awaited create/update where the run parks for approval will complete with a "waiting for input" warning instead of failing with "unknown building block status; provider may be out of date".
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
-# v0.23.1
+# v0.23.2
 
 FEATURES:
-- `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to `0` when omitted.
+- `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to the already set value when omitted, or `0` if there is none.
+
+# v0.23.1
 
 FIXES:
 - `meshstack_building_block`: Prepare for the upcoming `WAITING_FOR_APPROVAL` run status (an approval gate coming soon to meshStack — not available yet). Once meshStack starts returning it, an awaited create/update where the run parks for approval will complete with a "waiting for input" warning instead of failing with "unknown building block status; provider may be out of date".
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # v0.23.1
 
+FEATURES:
+- `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to `0` when omitted and is intentionally excluded from a version's `content_hash`, so reordering does not create a version change.
+
 FIXES:
 - `meshstack_building_block`: Prepare for the upcoming `WAITING_FOR_APPROVAL` run status (an approval gate coming soon to meshStack — not available yet). Once meshStack starts returning it, an awaited create/update where the run parks for approval will complete with a "waiting for input" warning instead of failing with "unknown building block status; provider may be out of date".
 - `meshstack_building_block_definition`: `Read` now derives `version_spec.draft` from the definition's actual latest version instead of retaining the prior state value. Previously, when the latest version was switched to a draft outside Terraform, refresh kept `draft = false`, so a `draft = false -> true` change created a redundant new version instead of reconciling the existing draft in place.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 FEATURES:
 - `meshstack_building_block_definition`: `version_spec.inputs` and `version_spec.outputs` now support an optional `display_order` (number) attribute that controls how inputs/outputs are arranged in meshPanel. It defaults to the already set value when omitted, or `0` if there is none.
 
+FIXES:
+- `meshstack_building_block`: A `content_hash` wired from a definition's computed `content_hash` no longer triggers a re-run when the value changed only because the hash-algorithm version changed (e.g. after upgrading the provider). Such version-only differences are now recognized and ignored, while genuine content changes still re-run. Setting an arbitrary (non-versioned) `content_hash` to force a manual re-run continues to work.
+
 # v0.23.1
 
 FIXES:
--- a/internal/provider/building_block_definition_resource.go
+++ b/internal/provider/building_block_definition_resource.go
@@ -218,6 +218,27 @@ func outputAssignmentTypes(outputs types.Map) map[string]string {
 	return result
 }
 
+func displayOrdersDiffer(a, b client.MeshBuildingBlockDefinitionVersionSpec) bool {
+	inputOrders := func(m map[string]*client.MeshBuildingBlockDefinitionInput) map[string]int64 {
+		orders := make(map[string]int64, len(m))
+		for key, input := range m {
+			if input != nil {
+				orders[key] = input.DisplayOrder
+			}
+		}
+		return orders
+	}
+	outputOrders := func(m map[string]client.MeshBuildingBlockDefinitionOutput) map[string]int64 {
+		orders := make(map[string]int64, len(m))
+		for key, output := range m {
+			orders[key] = output.DisplayOrder
+		}
+		return orders
+	}
+	return !maps.Equal(inputOrders(a.Inputs), inputOrders(b.Inputs)) ||
+		!maps.Equal(outputOrders(a.Outputs), outputOrders(b.Outputs))
+}
+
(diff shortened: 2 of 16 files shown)


internal/provider/building_block_definition_resource.go
```diff
@@ -217,6 +217,27 @@ func outputAssignmentTypes(outputs types.Map) map[string]string {
 	return result
 }
 
+func displayOrdersDiffer(a, b client.MeshBuildingBlockDefinitionVersionSpec) bool {
```
grubmeshi: f: hm, that's not quite the correct way of addressing this IMHO. the `versionContentHash` already has a part where it sets fields to constant/zero values. 

My proposal would be instead of adding this function (remove it), address it entirely [at this place](https://github.com/meshcloud/terraform-provider-meshstack/blob/dc72bff4802fdb6ad5a9094be1d6912d70482974/internal/provider/building_block_definition_resource_model.go#L323-L323). Loop over inputs/ouputs and set `displayOrder` to 0 no matter what the DTO sent, making the hash stable against any changes in there.

It might require making the `DisplayOrder` a `*int64`. I'm not sure of map entries only get ignored by our hashing algo if it's a `nil` pointer or also if it's a zero value (`0` in case for `int64`)

internal/provider/building_block_definition_resource_schema.go
```diff
@@ -185,6 +193,13 @@ func (r *buildingBlockDefinitionResource) Schema(_ context.Context, _ resource.S
 				stringvalidator.OneOf(client.MeshBuildingBlockOutputIOTypes.Strings()...),
 			},
 		},
+		"display_order": schema.Int64Attribute{
+			MarkdownDescription: "Numeric value controlling the display ordering of this output in meshPanel. " +
+				"Used only for arranging outputs; it does not affect the version's content hash. Defaults to `0` when omitted.",
```
grubmeshi: d: sounds a bit AI sloppy (also fix likewise for input above), maybe:

```suggestion
			MarkdownDescription: "Controls the display ordering of this output in meshStack UI. " +
				"It does not affect the version's content hash. Defaults to `0` when omitted. Values can be non-sequential to support groups, such as `01`, `02`, `03`, ..., `10`, `11`, `12`, ..., across all inputs. Duplicate values are resolved by alphanumeric sorting of keys as a fallback.",
```

also, is `meshPanel` the official product name? I'd have called it "meshStack UI" 🤔

grubmeshi (changes requested):
Have a look at my proposal of implementing it differently. I think only addressing this inside the resource is not enough and it must be handled inside `versionContentHash` itself.

internal/provider/building_block_definition_version_content_hash.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+
+	"github.com/meshcloud/terraform-provider-meshstack/client"
+	"github.com/meshcloud/terraform-provider-meshstack/internal/util/hash"
+)
+
+const (
+	errorHashVersion   = "err"
```
grubmeshi: f: this looks really weird, usually errors in Go implement the `error` as types or are `var errXXX = errors.New(...)`, but not `const`? comply with Go idioms or at least explain why that's not possible here.

internal/provider/building_block_definition_version_content_hash.go
```diff
@@ -0,0 +1,159 @@
...
+	currentHashVersion = "v2"
+)
+
+// represents a content hash of a building block definition version, which is used to detect changes in the version_spec.
+// because the version_spec is subject to change w.r.t. new fields etc., the hash is versioned
+// this means we can only meaningfully compare hashes of the same version
+// use func indicatesChangesTowardsHash(hashStr string) to detect whether an instance of BuildingBlockDefinitionVersionContentHash
+// differs from a previously calculated hash string representation.
+type BuildingBlockDefinitionVersionContentHash struct {
+	hashVersion string
```
grubmeshi: d:
```suggestion
	hashVersion int
```
maybe parse this to a number making version comparisions much safer/easier?

internal/provider/building_block_definition_version_content_hash.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+
+	"github.com/meshcloud/terraform-provider-meshstack/client"
+	"github.com/meshcloud/terraform-provider-meshstack/internal/util/hash"
+)
+
+const (
+	errorHashVersion   = "err"
+	currentHashVersion = "v2"
```
grubmeshi: d: (see other comment at `hashVersion` field
```suggestion
	currentHashVersion = 2
```

internal/provider/building_block_definition_version_content_hash.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(parts) != 2 {
+		h.hashVersion = errorHashVersion
+		return
+	}
+
+	h.hashVersion = parts[0]
+	h.hashValue = parts[1]
+}
+
+func errorHash() BuildingBlockDefinitionVersionContentHash {
```
grubmeshi: f: I really don't link using `errorHashVersion` as a way to indicate errors, why not just return proper Go `error` implementing instances with also good description what is going on, potentially using machinery like `errors.Is` to react on certain errors at the calling site?

internal/provider/building_block_definition_version_content_hash_test.go
```diff
@@ -0,0 +1,137 @@
...
+	assert.Equal(t, hashSame, irrelevant.compareToStored(stored))
+	assert.Equal(t, hashDifferent, relevant.compareToStored(stored))
+}
+
+// Mirrors the released-version immutability guard's fail-safe: when the stored hash is from a legacy /
+// unknown algorithm version (e.g. written by an older provider and never refreshed), it is incomparable
+// to the current-version hash. The guard then recomputes the released spec from state AT THE CURRENT
+// version so the comparison becomes meaningful again — a genuine change is detected (would reject) and
+// an unchanged spec is recognized (no-op), instead of the old silent "no change" guess.
+func Test_contentHash_incomparableStoredHash_resolvedByRecomputeAtCurrentVersion(t *testing.T) {
```
grubmeshi, line 90: d: I think this logic is vital and should be part of the type `BuildingBlockDefinitionVersionContentHash` and it is almost done by `compareToStored`. The other comment concerning content hash change detection to determine BB re-run could also benefit from revisiting the design again I suppose.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # v0.23.1
 
+FEATURES:
```
grubmeshi, line 3: f: `v0.23.1` is already released, you need to add another `v0.23.2` section.

internal/provider/building_block_definition_resource_model_test.go
```diff
@@ -62,7 +62,7 @@ func Test_versionContentHash_plaintextSecret(t *testing.T) {
 	var versionSpec client.MeshBuildingBlockDefinitionVersionSpec
 	require.NoError(t, json.Unmarshal(versionSpecPlaintextSecretJson, &versionSpec))
 	var diags diag.Diagnostics
-	versionContentHash(versionSpec, &diags)
+	calculateBuildingBlockDefinitionVersionContentHash(versionSpec, &diags)
```
grubmeshi: f: those tests for the old `versionContentHash` should also move to the separate file you've created in this PR (which is a good idea!)

internal/provider/building_block_definition_resource_model_test.go
```diff
@@ -30,9 +30,9 @@ var (
 func Test_versionContentHash(t *testing.T) {
 	// If constant values below are required to change, you need a good reason and consider backwards compatibility!
 	const (
-		hashWhichShouldNeverChange1 = "v1:674c77c28e4eb4cec99b9f1e73ad11b520a367da416ff3fa90d5e5426e09befc"
-		hashWhichShouldNeverChange2 = "v1:020cd7c032ff6ce05a02873fd319e7b7206896fa415904791836c933f0a239ee"
-		hashWhichShouldNeverChange3 = "v1:2b992e234316baa08d1f4d017f57379ddfb45f39f3ae7b148079322938dfd8e0"
+		hashWhichShouldNeverChange1 = "djI6Njc0Yzc3YzI4ZTRlYjRjZWM5OWI5ZjFlNzNhZDExYjUyMGEzNjdkYTQxNmZmM2ZhOTBkNWU1NDI2ZTA5YmVmYw=="
```
grubmeshi, line 32: n: thank you for making this opaque. Remember to close issue #209 then (use `Fixes #209` in commit).

internal/provider/building_block_definition_resource.go
```diff
@@ -490,21 +490,40 @@ func (r *buildingBlockDefinitionResource) Update(ctx context.Context, req resour
...
+		versionSpecDtoContentHash := calculateBuildingBlockDefinitionVersionContentHash(versionSpecDto, &resp.Diagnostics)
 		if resp.Diagnostics.HasError() {
 			return
 		}
-		if versionSpecDtoContentHash == state.VersionLatest.ContentHash.Get() {
-			// state is in draft=false (aka released), and there's no change in version_spec,
+
+		cmp := versionSpecDtoContentHash.compareToStored(state.VersionLatest.ContentHash.Get())
+		if cmp == hashIncomparable {
+			// The stored hash was produced by a different/legacy algorithm version (state written by an
```
grubmeshi, line 500: d: the new `meshstack_building_block` resource uses `content_hash` as an input (via the read-only `version*` BBD attributes) to determine if a BB needs to re-run (relevant when developing/testing a BBD draft version with a same workspace BB for platform operators). I expect that this logic is also there, see also other comment proposing to move that logic closer to the type `BuildingBlockDefinitionVersionContentHash`, so it can be easily shared.

I also wondered if we need the catch here that BB re-runs are actually not necessary when the content hash changes just because display_order is different, but this optiimization can be thought of later I suppose.

grubmeshi (changes requested):
Please address my review comments, in general, the approach is sound I think.

internal/provider/building_block_definition_version_content_hash_test.go
j0g3sc: I disagree. This is a combination of 1) reading json version spec and 2) calculate the hashsum.
2) is already part of BuildingBlockDefinitionVersionContentHash but 1) should be done beforehand and not by the hash algorithm. For testing this is fine, but not in the production code.

grubmeshi (approved):

PR #220, opened 2026-07-06
feat: capture a failed composed child's run log via the run token

grubmeshi:
# Capture a failed composed child's run log via the run token

Pairs with the meshfed-release PR on the identically-named branch `feature/composed-run-failure-logs`.

## Why

When a composition building block (e.g. a starter kit) waits on a child building block it created and that
child fails, `addRunFailureDiagnostics` fetches the child run's logs and adds the failing step as an error, so
the tf-block-runner streams the real error into the parent run's log. Fetched with the consumer's workspace
credentials, a child with `run_transparency = false` returns no `systemMessage`, so the parent log only carried
a generic "run failed". meshStack injects a privileged, run-scoped token (`MESHSTACK_RUN_TOKEN`) into such runs
that can read the composition's own children's runs.

## Change

- `provider.go`: read `MESHSTACK_RUN_TOKEN` (env-only — it is set per run by the platform, never in HCL); when
  present, build a run-token-authorized run-log client.
- `client.go`: `NewBuildingBlockRunClientWithAuth` builds a standalone `MeshBuildingBlockRunClient` with its own
  authorization (skipping the version check the main client already did). New `Client.BuildingBlockRunWithRunToken`
  field, nil unless the token is set.
- `building_block_resource.go`: `addRunFailureDiagnostics` fetches the failed run's logs with the run-token
  client when available, else the workspace client. Everything else (CRUD, polling) keeps the workspace
  credentials.

The child run uuid the provider polls comes from `bb.Status.LatestRunUuid`, which the paired backend change
now exposes to the block owner regardless of the child's run transparency.

Additive on a `-preview` API — not breaking, no `TerraformProviderVersionRequirements.kt` coordination needed.

## Tests

- Unit (`building_block_resource_await_test.go`): `addRunFailureDiagnostics` uses the **run-token** client when
  configured (and not the workspace client), and falls back to the workspace client when it is not; the fetched
  system message reaches the diagnostics. Hermetic, CI-runnable.
- End-to-end is verified by hand with a local scratch demo (a composition against a local stack with one child
  forced to fail at `run_transparency = false`); it is not automated because CI runs the tf-block-runner as a
  service container that pulls the released provider from the registry, so it cannot exercise a dev-built
  provider. Wiring the dev provider into the local runner requires a `dev_overrides` block in `$HOME/.terraformrc`
  (the runner strips `TF_CLI_CONFIG_FILE`).

`task test` and `task lint` green. `CHANGELOG.md`: new `v0.23.1` FEATURES entry.

## Cross-repo PR

- meshcloud/meshfed-release#10286 — backend changes (A: expose run identifiers to the block owner; B: run token for sync API runs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 FEATURES:
 - `meshstack_building_block`: A composition now captures a failed child building block's full run log, using the run-scoped `MESHSTACK_RUN_TOKEN` when meshStack injects it, even if the child has run transparency disabled.
 
+FIXES:
+- `meshstack_building_block_definition`: A definition with an empty `version_spec.inputs = {}` no longer fails with "Provider produced inconsistent result after apply" (null vs empty map). The applied and refreshed values now preserve the configured shape.
+
 # v0.23.0
 
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v0.23.1
+
+FEATURES:
+- `meshstack_building_block`: A composition now captures a failed child building block's full run log, using the run-scoped `MESHSTACK_RUN_TOKEN` when meshStack injects it, even if the child has run transparency disabled.
+
 # v0.23.0
 
 FEATURES:
--- a/client/client.go
+++ b/client/client.go
@@ -41,6 +41,10 @@ type Client struct {
 	Workspace                      MeshWorkspaceClient
 	WorkspaceGroupBinding          MeshWorkspaceGroupBindingClient
 	WorkspaceUserBinding           MeshWorkspaceUserBindingClient
+
+	// BuildingBlockRunWithRunToken reads run logs authenticated with the privileged run token, letting a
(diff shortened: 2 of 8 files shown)


grubmeshi:
Superseded — closing and dropping this branch.

The better approach is handled entirely in the meshStack backend: a composition's ephemeral run key is treated as a platform operator and granted access to a failed child block's run logs even when `run_transparency = false`. This follows the ephemeral-run-key operator-elevation pattern now on `develop` (meshcloud/meshfed-release@b4b1bac, @d1dc188), where starterkit / composed BB runs are recognized as platform operators with somewhat elevated permissions.

With the backend granting log access directly to the run key, the provider no longer needs to fetch a failed child's logs via `MESHSTACK_RUN_TOKEN`, so these provider changes (and the accompanying 'preserve empty vs null inputs' fix) are dropped.

Follow-up (tracked in the ClickUp CR ticket): narrow the elevation from workspace scope down to the composition block's UUID / ownership.

PR #225, opened 2026-07-07
fix(building_block_definition): reflect external draft state on Read

grubmeshi:
## What

`meshstack_building_block_definition`'s `Read` derived `version_spec.draft` from the **prior state value** rather than the definition's actual latest version. When state held `draft = false` but the latest version had been switched to a draft **outside Terraform**, refresh kept `draft = false` and only warned. A subsequent `draft = false -> true` change then took the "create a new draft version" path and **created a redundant new version** (e.g. v14 → v15) instead of reconciling the existing draft in place.

## Fix

Derive `draft` from the actual latest version on refresh via `deriveDraftFromLatestVersion(...)` — the same helper the `meshstack_building_block_definitions` data source already uses. The release-pending-approval warning stays in `Create`/`Update`, which pass the *known desired* draft.

## Verification

Reproduced against a real backend with a BBD whose latest version was an external draft while state had `draft = false`:

- **Before:** plan shows `draft = false -> true`, `version_number 14 -> 15`, a new `DRAFT` version appended.
- **After:** `draft` correctly stays `true`; the existing draft is updated in place — `Plan: 0 to add, 1 to change, 0 to destroy`.

Unit tests (`go test -run BuildingBlockDefinition ./internal/provider/`) pass.


grubmeshi:
Rebased into a stack for a single fast-forward merge to `main`: `main → #223 → #224 → #225`. This branch's tip (`d09192e`) contains all three PRs, so CI here exercises the combined result; merging pushes the stack tip to `main`, closing all three.

PR #228, opened 2026-07-07
fix: track only user-declared tags on taggable resources

tfelix:
## Problem

Resources whose tag schema contains a **restricted tag with a default value** — or simply any
undeclared tag property — could not be managed cleanly via Terraform:

- `meshstack_landingzone` — `terraform apply` crashed on create with *"Provider produced inconsistent result after apply"*.
- `meshstack_project` — apply succeeded, but every subsequent `terraform plan` showed drift, trying to remove a tag the caller may not be permitted to manage.

## Root cause

On the meshObject API create path, meshStack returns a tag map that is a **superset** of what the
caller sent: it fills in an entry for every defined tag property (empty list for unset ones) and,
for `meshProject` / `meshLandingZone` / `meshBuildingBlockDefinition`, injects defaults for
restricted tags. `tags` is an `Optional + Computed` attribute, so writing that superset into it
violated the plugin framework's plan/apply consistency rule (the landing zone crash), and mirroring
it on `Read` produced perpetual drift (the project case). For restricted tags the caller may lack
permission to set them at all, so the documented "just set it to the default yourself" workaround
does not apply.

## Fix (commit 1 — `fix: track only user-declared tags on taggable resources`)

`tags` now tracks only the tags the user declares, across every taggable resource
(`meshstack_landingzone` / `_project` / `_workspace` / `_payment_method` / `_building_block_definition`):

- **Create / Update** persist the plan's tags instead of the API's superset.
- **Read** reconciles the API response down to the keys already tracked in state (`reconcileTrackedTags`), so server-injected entries never enter `tags` and never surface as drift.
- **Import** keeps the full set meshStack returns (no prior state to reconcile against), so a normal import still round-trips.

## Tag schema cleanup (commit 2 — `refactor: consolidate tag schema into a shared helper; model tags as list`)

- A shared `tagsAttribute(tagsOptions{Kind, Output, Restricted})` helper (in `schema_utils.go`, mirroring the `meshRef` helpers) replaces the hand-rolled `tags` blocks on all taggable resources and data sources. `reconcileTrackedTags` / `reconcileTags` moved to `tags_reconcile.go` (reconciliation, not schema).
- **Tag values are now `list(string)` everywhere.** A backend check confirmed tag values are ordered lists, never sets (`Taggable.kt` = `Map<String, List<String>>`; `TagService` never dedupes/sorts/sets; `multiSelect` is an ordered list; the panel stores an ordered array). The `meshstack_landingzone` resource was the lone `set(string)` (shipped in v0.24.0) — a `set` would silently dedupe/reorder values the backend preserves. It becomes `list` with a schema **v0 → v1 `UpgradeState`** that migrates existing state in place (value-preserving; HCL unchanged).
- The `Restricted` flag (which surfaces the "restricted-tag defaults are not tracked" doc note) is set only for the kinds whose **meshObject API** create path actually injects restricted defaults — `meshProject`, `meshLandingZone`, `meshBuildingBlockDefinition`. `meshWorkspace` injects them only via the panel registration path (not the API), and `meshPaymentMethod` never does, so both use the generic superset note.

## Tests

- Per-resource acceptance subtests against the real backend: `restricted_default_tag` (project / landing zone / bbd) and `superset_tag_reconciliation` (workspace / payment method); plus landing-zone `only_restricted_injected_no_declared_tags` (the original crash repro) and `declared_restricted_tag_kept`.
- A pure `reconcileTags` unit test (`TestReconcileTags`) covering the drop/keep matrix (multiple tracked keys, declared-restricted kept vs injected dropped, only-restricted → empty).

## Docs & changelog

- `CHANGELOG.md`: `v0.24.1` `FIXES:` entries for the tags fix and the landing-zone `set`→`list` correction.
- Regenerated resource/data-source docs; every `tags` attribute now shares one description.


internal/clientmock/mock_landingzone.go
```diff
@@ -9,11 +9,23 @@ import (
...
 func (m MeshLandingZoneClient) Read(_ context.Context, name string) (*client.MeshLandingZone, error) {
 	v, _ := m.Store.Get(name)
-	return v, nil
+	if v == nil {
+		return nil, nil
+	}
+	// Return a copy so callers that reassign fields (e.g. tags) can't mutate the stored object,
+	// mirroring a real backend that serializes a fresh object on every request.
+	cp := *v
+	return &cp, nil
```
grubmeshi: f: use `new`? (see `modern-go` skill)


```suggestion
	return new(*v), nil
```

internal/clientmock/mock_landingzone.go
```diff
@@ -25,8 +37,10 @@ func (m MeshLandingZoneClient) Create(_ context.Context, landingZone *client.Mes
 			Restricted: false,
 		},
 	}
+	created.Metadata.Tags = mergeInjectedTags(created.Metadata.Tags, m.InjectTagsOnCreate)
 	m.Store.Set(landingZone.Metadata.Name, created)
-	return created, nil
+	cp := *created
+	return &cp, nil
```
grubmeshi: f: use `new`?

internal/clientmock/mock_landingzone.go
```diff
@@ -9,11 +9,23 @@ import (
...
+	InjectTagsOnCreate map[string][]string
 }
 
 func (m MeshLandingZoneClient) Read(_ context.Context, name string) (*client.MeshLandingZone, error) {
 	v, _ := m.Store.Get(name)
-	return v, nil
+	if v == nil {
+		return nil, nil
+	}
+	// Return a copy so callers that reassign fields (e.g. tags) can't mutate the stored object,
```
grubmeshi: f: make comments more succinct, also check other comments. only explain surprising behavior and not what the code already expresses!

internal/provider/tags_restricted_default_test.go
```diff
@@ -0,0 +1,103 @@
...
+var injectedServerTags = map[string][]string{
+	"restricted-default": {"injected"},
+	"unset-property":     {},
+}
+
+// TestAccLandingZoneRestrictedDefaultTag guards against "Provider produced inconsistent result after
+// apply" (and follow-up drift) when the backend injects restricted-tag defaults on create. See
+// TagService.determineTags / LandingZoneMaintenanceService.applyLandingZoneTagsOnCreation on the
+// backend. Mock-only: the injection is simulated by the mock client.
+func TestAccLandingZoneRestrictedDefaultTag(t *testing.T) {
```
grubmeshi: f: do not add mock only tests but see to get this working with a real locally running backend. I know existing tests still work, but I don't see much value in mock tests here as it's really the backend behavior which needs to be covered. 

To this end, a real backend acceptance test sets up actual restricted tags using the `meshstack_tag_definition` resource in a test support file. Also feel free to use the scratch testing skill to experiment locally to find a good solution. We already go a long way of proper test setup, see e.g. the BB acceptance tests with tenant level (Setting up project, LZ, tenants, BBD, workspace just to run a BB properly locally).

Also, convention is that no extra go test files exists but acceptance tests are structured into subtests (see skill). My expectation is that each resource (project / landingzone) has its own added subtest, even if that means duplicating test support files. Also, I don't see a strong reason to use `IsMockClientTest` which should only be used exceptionally (isn't there a comment explaining that?)

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v0.23.2
+
+FIXES:
+- `meshstack_landingzone` / `meshstack_project`: Fixed *"Provider produced inconsistent result after apply"* (landing zone) and perpetual after-apply drift (project) for resources whose tag schema contains restricted tags with a default value. meshStack fills such defaults in on creation — and returns an entry for every defined tag property — so the response is a superset of what the caller sent, which the caller may not be permitted to manage. The `tags` attribute now tracks only the tags you declare; server-injected restricted defaults are accepted silently and no longer surface as diffs. On import, `tags` still reflects the full set meshStack returns.
```
grubmeshi: d: feature wise, are `landingzone` and `project` the only tags-resources where such "restricted" tags can appear? If not, please add this support for all resources where it might appear.

also, consider adding a new read only attribute `all_tags` (similar to the `all_inputs` output of BB resource which deals with similar, but even more complicated situations) which exposes the "immutable" restricted tags along with the tags managed by the resource.

grubmeshi (changes requested):
General idea sounds valid, however, we need better test support and also see remarks on feature in general (on changelog entry).

CHANGELOG.md
tfelix: Okay yeah there is actually another resource with a similiar issue. I seemingly misunderstood the AI but after a bit more research there is the bbd. payment method and the workspace. All have a similiar issue and needs to be fixed.

internal/clientmock/mock_landingzone.go
tfelix: fyi I fixed it but this causes issues with VS codes go language support. 

image

While it compiles just fine. But such "no-errors" are still confusing and could hide actual errors.
I suggest to stick to Go conventions which are supported by the IDE the majority of developers use to stay compatible.

internal/clientmock/mock_landingzone.go
tfelix: same as above.

internal/clientmock/mock_landingzone.go
tfelix: f

internal/provider/tags_restricted_default_test.go
tfelix: "The wrinkle: the acceptance-testing skill's lock-step policy actually lists "defaults the backend materializes" as one of the few sanctioned reasons to gate a whole subtest on IsMockClientTest() — which is exactly this scenario."

So this explicit use case was mentioned as an acceptable way for using mocks. But I think it probably would also work against a real backend.

I restructured the tests now to fit better into your suggested solution and pointed them at the real backend.

CHANGELOG.md
tfelix: Good idea, but I'd like to keep it out of scope for this PR — this one is a bugfix for the crash/drift on tags callers already declare, not a new feature. `all_tags` (mirroring `all_inputs` on the BB resource) is a genuine feature addition with its own schema/doc/test surface across both resources, so I'd rather track it as a follow-up than expand this fix's blast radius.

tfelix:
Maybe it make sense to go face 2 face over the commits because I am not familiar especially with the testing code here.

internal/clientmock/mock_landingzone.go
grubmeshi: Ah c'mon use a proper IDE *whisper* GoLand *whisper* :P

internal/provider/workspace_resource.go
```diff
@@ -177,6 +183,24 @@ func (r *workspaceResource) Read(ctx context.Context, req resource.ReadRequest,
...
+	// verbatim would surface as drift. On import there is no prior state (tags is null); we keep the
+	// full set so a normal import round-trips.
+	var priorTags types.Map
+	resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("metadata").AtName("tags"), &priorTags)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if !priorTags.IsNull() {
+		var tracked map[string][]string
+		resp.Diagnostics.Append(priorTags.ElementsAs(ctx, &tracked, false)...)
```
grubmeshi: f: remove? why is this put into diags?

internal/provider/schema_utils.go
```diff
@@ -33,6 +33,21 @@ func emptySetDefault(nested schema.NestedAttributeObject) defaults.Set {
 	return setdefault.StaticValue(types.SetValueMust(nestedObjectToObjectType(nested), nil))
 }
 
+// reconcileTrackedTags restricts apiTags to the keys already tracked in priorTags. The meshObject
+// API returns a superset of the tags a caller sent — an entry for every schema property (empty list
+// for unset ones) plus injected restricted-tag defaults — which the caller may be unable to manage.
+// Keeping only the previously tracked keys prevents those server-side additions from entering the
+// user-managed `tags` attribute and producing spurious drift on the next plan.
+func reconcileTrackedTags(priorTags, apiTags map[string][]string) map[string][]string {
```
grubmeshi: I think one can move also the state access inside this helper function saving a redundant
```
var priorTags types.Map
	resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("metadata").AtName("tags"), &priorTags)...)
```
at call site, maybe?

error handling is a bit annoying though, but pass in diags as pointer (see other methods like SetFromClientDto etc.)

internal/provider/workspace_resource_test.go
```diff
@@ -11,11 +11,43 @@ import (
...
+		// undeclared ones); the mock has no tag-schema logic and can't reproduce it. See the lock-step
+		// policy in the acceptance-testing skill.
+		if IsMockClientTest() {
+			t.Skip("relies on the backend returning an entry for every defined tag property")
+		}
+
+		config, wsAddr := testconfig.Workspace(t)
+		// A second tag definition the workspace does not declare: the backend still returns it as an
+		// empty list, so the fix must reconcile it away instead of surfacing it as drift.
+		undeclaredTag, _, _ := testconfig.TagDefinition(t, client.MeshObjectKind.Workspace)
```
grubmeshi, line 31: f: why is this not using restricted tag? not supported for workspaces?

internal/provider/project_resource_test.go
```diff
@@ -11,11 +11,46 @@ import (
...
 	"github.com/meshcloud/terraform-provider-meshstack/internal/provider/acctest/xknownvalue"
 )
 
 func TestAccProject(t *testing.T) {
+	t.Run("restricted_default_tag", func(t *testing.T) {
+		// Backend-materialized default: the mock has no tag-restriction business logic, so it can't
+		// reproduce TagService.determineTags injecting a restricted tag's default on create. See the
+		// lock-step policy in the acceptance-testing skill.
+		if IsMockClientTest() {
+			t.Skip("relies on the backend injecting a restricted tag's default value on create")
```
grubmeshi, line 25: f: still there are changs left in mock client but now all added tests run with actual backend only. so either restore "dummy" mock support but simulating restricted tags (needs store sharing in mock so it detects the now addtional tag definitions to return them, which is actually not that hard to mock), or remove all changes in `clientmock` package.

internal/provider/acctest/testconfig/build_tag_definition.go
```diff
@@ -22,3 +22,26 @@ func TagDefinition(t *testing.T, targetKind string) (config Config, tagDefinitio
...
+	t.Helper()
+	keySuffix := acctest.RandString(8)
+	tagKey = "test-key-" + keySuffix
+	config = Resource{Name: "tag_definition"}.Config(t).WithFirstBlock(
+		RenameKey("example_"+keySuffix),
+		ExtractAddress(&tagDefinitionAddr),
+		Descend("spec")(
+			Descend("target_kind")(SetString(targetKind)),
+			Descend("key")(SetString(tagKey)),
+			Descend("restricted")(SetRawExpr("true")),
```
grubmeshi: d:
```suggestion
			Descend("restricted")(SetBool("true")),
```
is embarrasingly not provided? do you mind adding it (and reducing the usage as SetRawExpr)? Maybe even `SetTrue()` might be warranted depending on how often it could be used.

internal/provider/acctest/testconfig/build_tag_definition.go
```diff
@@ -22,3 +22,26 @@ func TagDefinition(t *testing.T, targetKind string) (config Config, tagDefinitio
...
+			Descend("value_type", "email")(
+				RenameKey("string"),
```
grubmeshi: f: looks like instead of cumbersomely adapting an existing example tag definition config, better define a test support file instead with fits better it's purpose (probably everything here can be removed then, except setting dynmaic values such as random suffixes etc)?

in particular spotting `RenameKey` here is weird and that method should only be used rarely (add this in godoc for that method)!

grubmeshi (approved):
Looks much better, see some more comments. Let's also huddle tomorrow I'd say!

internal/provider/schema_utils.go
tfelix: Done — `reconcileTrackedTags(ctx, state, tagsPath, apiTags, &diags)` now reads the prior tags from state itself and appends diagnostics via the pointer, so the redundant `var priorTags types.Map` / `GetAttribute` block is gone and the five call sites (incl. the BBD one) collapse to a single line. I also extracted a pure `reconcileTags(tracked, apiTags)` core and unit-tested it (`TestReconcileTags`).

internal/provider/workspace_resource.go
tfelix: Resolved by the `reconcileTrackedTags` refactor — the `ElementsAs` conversion (and its diagnostics) now lives inside the helper, not at the call site.

internal/provider/workspace_resource_test.go
tfelix: Workspaces *do* support restricted tags — the backend injects restricted-tag defaults on create via `WorkspaceRegistrationController.applyDefaultTags` (same `restrictedDefaultsAsSerialized()` path as project/landing zone). This subtest deliberately covers the broader **superset** case instead: the API returns an empty-list entry for every defined tag property, even undeclared ones, which is the more general reconciliation the fix has to handle. Happy to add a dedicated restricted-default workspace subtest too if you'd like both paths covered explicitly.

internal/provider/acctest/testconfig/build_tag_definition.go
tfelix: Added `SetBool(bool)` to `config_expr.go`. `restricted = true` now comes from the new test-support file, so the `SetRawExpr("true")` here is gone. I left `SetTrue` out for now — `SetBool(true)` reads clearly enough — but easy to add if it starts to proliferate.

internal/provider/acctest/testconfig/build_tag_definition.go
tfelix: Switched to purpose-built `test-support_tag.tf` / `test-support_restricted-tag.tf` (both declare `value_type = { string = ... }` directly), so the `email`→`string` `RenameKey` hack is gone; the builders now only set dynamic values (target_kind, randomized key, default_value). `RenameKey` stays only to give each block a unique label per run — needed because builders like `ProjectAndWorkspace` already contribute a tag definition — and I documented that as its rare, sanctioned use in the `RenameKey` godoc.

internal/provider/project_resource_test.go
tfelix: Went with removing the `clientmock` changes: the added tests run against the real backend only, and simulating restricted-default injection in the mock would duplicate backend business logic (risking drift / false confidence) — which was your earlier point about mock tests here. To keep a fast unit-level guard on the actual fix logic without a live backend, I extracted the pure `reconcileTags` core and added `TestReconcileTags`. Happy to go the other way (faithful mock via a shared tag-definition store, as you sketched) if you'd rather have mock-mode coverage — let's settle it when we sync.

internal/provider/workspace_resource.go
grubmeshi: Addressed by the `reconcileTrackedTags` refactor — the `ElementsAs` conversion and its diagnostics now live inside the helper, so there is no diags handling at this call site anymore. Rebased onto current `main` (v0.24.1). Resolving.

internal/provider/schema_utils.go
grubmeshi: Done — `reconcileTrackedTags(ctx, state, tagsPath, apiTags, &diags)` reads the prior tags from state itself and appends diagnostics via the pointer, so the redundant `var priorTags types.Map` / `GetAttribute` block is gone and the five call sites collapse to one line. A pure `reconcileTags` core is extracted and unit-tested (`TestReconcileTags`). Resolving.

internal/provider/workspace_resource_test.go
grubmeshi: Added a dedicated `TestAccWorkspace/restricted_default_tag` subtest (mirrors project/landingzone/bbd): workspaces do inject restricted-tag defaults via `WorkspaceRegistrationController.applyDefaultTags`, so the restricted path is now covered on workspaces too; the existing `superset_tag_reconciliation` subtest stays for the broader empty-list case. Also broadened coverage on landing zone (`only_restricted_injected_no_declared_tags`, `declared_restricted_tag_kept`) and in `TestReconcileTags`. Resolving.

internal/provider/acctest/testconfig/build_tag_definition.go
grubmeshi: Switched to purpose-built `test-support_tag.tf` / `test-support_restricted-tag.tf` so the builders only set dynamic values; the `email`→`string` `RenameKey` hack is gone. `RenameKey` now carries a godoc marking it as a rare, sanctioned use (per-run unique block labels). Resolving.

PR #232, opened 2026-07-13
refactor: make display_order again part of the definition version content hash

j0g3sc:


grubmeshi (approved):
LGTM, that was essentially all that was missing, indeed.

PR #235, opened 2026-07-14
refactor: consolidate meshObject reference handling behind a single meshRef helper

grubmeshi:
Consolidates the provider's reference-schema handling so every meshObject reference is built the same way, and brings the last outlier (`building_block_definition_version_ref`) in line by giving it a `kind`.

## What changed

- **One builder, two constructors.** `meshRefByUuid` / `meshRefByName` take a `meshRefOptions` struct with three behaviour flags — `Output`, `OptionalComputed` and `InSet`; the zero value is a required input (block **and** identifier `Required`). This replaces the previous `meshRef(kind, refId, desc, ...opts)` signature and its `asOutput()` / `required()` / `optionalComputed()` / `requiredId()` functional options. Every call site is repointed.
- **Only set-nested identifiers are unknown-tolerant (`InSet`).** A ref hashed as an opaque set element — nested in a `SetNestedAttribute` object (e.g. `project_role_ref` in a role-mapping set) or used as a set's element type (e.g. `mandatory_building_block_refs`) — cannot use a plain `Required` identifier: the element collapses to a wholly-unknown value at plan (Terraform hashes set elements by whole value) and `Required` then fails with *"Missing Configuration for Required Attribute"*. `InSet` keeps such identifiers `Optional+Computed` with an `AlsoRequires` guard that enforces presence while tolerating an identifier that resolves after apply (a computed `.ref`, a random suffix, …), and documents this in the attribute description. **Standalone and list-nested refs now use a plain `Required` identifier** — clearer schema and docs, matching the backend. Documented inline in `schema_utils.go` and `AGENTS.md`.
- **Output refs keep `kind` known at plan time.** A `refOutputKind` plan modifier fills a still-unknown output block with its constant `kind` and an unknown identifier, instead of planning the whole object as "known after apply". Verified by a `PreApply` plan check in `TestAccWorkspace`. Data sources have no plan-modifier hook, so this is resource-only; data-source refs resolve `kind` at their plan-time read.
- **Platform `aws`/`gcp` role mappings are now sets, not lists.** The backend stores AWS (`aws_sso`, `aws_identity_store`) and GCP platform role mappings in a map keyed by the referenced meshProjectRole, so their order is neither meaningful nor preserved — modeling them as ordered `ListNestedAttribute`s produced a permanent no-op plan diff whenever the backend returned them in a different order. They are now `SetNestedAttribute`s (matching azure/openshift and all landingzone role mappings); the client DTO fields are retyped to the generic `client/types.Set[T]` so the DTO→model converter emits set values. This also resolves the review note that `project_role_ref.name` looked wrongly optional: as a set element the lenient (`InSet`) identifier is correct.
- **`building_block_definition_version_ref` gains a `kind`** (computed, always `meshBuildingBlockDefinitionVersion`) across all schema sites and the definition's `version_latest` / `version_latest_release` / `versions` outputs. The client struct gains a `Kind` field; Read/Create force the fixed discriminator so it round-trips even if the backend omits it.

## Deliberately left bespoke

- `target_ref` (discriminated uuid-or-name) and `building_block_definition_version_ref` (extra `content_hash` on the v3 path) carry extra fields, so they stay bespoke — the latter now simply also carries a kind. Landingzone `platform_ref` stays bespoke (its `uuid` uses `RequiresReplace`).

## User-facing impact (see CHANGELOG under pending v0.23.3)

- `building_block_definition_version_ref` now carries `kind` (FEATURES).
- `meshstack_platform` / `meshstack_landingzone`: `project_role_ref` in role mappings now has an optional `kind` (defaulting) instead of read-only and no longer requires a known-literal identifier, so a referenced resource's computed `ref` can be assigned (FIXES).
- `meshstack_platform`: `aws`/`gcp` role mappings are now unordered sets instead of ordered lists, ending the permanent no-op plan diff caused by backend reordering (FIXES). HCL block syntax is unchanged; a one-time no-op diff may appear on the first plan after upgrading.
- Standalone/list ref identifiers now render as `Required` in the docs (previously `Optional`); enforcement is unchanged.

## Validation

- `go build`, `task lint` (0 issues), `task generate` (docs regenerated, no drift), full unit suite (`go test ./...`, mock mode) all pass.
- New acceptance coverage: platform resource asserts `aws`/`gcp` role mappings resolve `project_role_ref` as a set; the platform data source asserts a read-only `project_role_ref` inside `aws_role_mappings` (previously unasserted — only openshift role mappings were covered).

## Commit structure

1. `refactor: consolidate meshObject reference handling behind a single meshRef helper` — the `meshRef` builder incl. the `InSet` flag and all call sites.
2. `refactor: consolidate client ref DTOs into shared NamedRef/UuidRef` — client-package alignment (`client/refs.go`); wire format & tfsdk schemas unchanged, no user-facing change.
3. `refactor: model platform aws/gcp role mappings as sets` — schema + `types.Set[T]` DTO retype.
4. `test: cover platform aws/gcp role mapping refs on resource and data source`.


internal/provider/building_block_v2_data_source.go
```diff
@@ -68,6 +68,10 @@ func (d *buildingBlockV2DataSource) Schema(ctx context.Context, req datasource.S
 								MarkdownDescription: "UUID of the building block definition.",
 								Computed:            true,
 							},
+							"kind": schema.StringAttribute{
+								MarkdownDescription: "meshObject type, always `" + client.MeshObjectKind.BuildingBlockDefinitionVersion + "`.",
```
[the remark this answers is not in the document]
grubmeshi: Fixed in f0df438: this block is the version ref, so both the block description and the `uuid` now say "building block definition version", matching the `meshstack_building_blocks` data source wording.

docs/data-sources/platform.md
```diff
@@ -351,22 +351,19 @@ Read-Only:
 <a id="nestedatt--spec--config--aws--replication--aws_identity_store--aws_role_mappings"></a>
 ### Nested Schema for `spec.config.aws.replication.aws_identity_store.aws_role_mappings`
 
-Required:
-
-- `project_role_ref` (Attributes) the meshProject role (see [below for nested schema](#nestedatt--spec--config--aws--replication--aws_identity_store--aws_role_mappings--project_role_ref))
```
grubmeshi, line 356: d: ah we're a datasource here and that is actually sth returned from the backend? so that's actually sth we fixed? is this covered in acc tests (if not, ensure it is).

docs/resources/platform.md
```diff
@@ -2035,18 +2023,15 @@ Read-Only:
...
 
-Required:
-
-- `name` (String) The identifier of the meshProjectRole
-
-Read-Only:
+Optional:
 
 - `kind` (String) meshObject type, always `meshProjectRole`.
+- `name` (String) Named identifier (`metadata.name`) of `meshProjectRole`.
```
grubmeshi: f: I think this change is wrong though and double check with backend and acc test

docs/resources/landingzone.md
```diff
@@ -187,18 +184,15 @@ Required:
...
 - `policies` (Set of String) List of policies associated with this role mapping
-- `project_role_ref` (Attributes) the meshProject role (see [below for nested schema](#nestedatt--spec--platform_properties--aws--aws_role_mappings--project_role_ref))
+- `project_role_ref` (Attributes) Reference to the meshProjectRole. (see [below for nested schema](#nestedatt--spec--platform_properties--aws--aws_role_mappings--project_role_ref))
 
 <a id="nestedatt--spec--platform_properties--aws--aws_role_mappings--project_role_ref"></a>
 ### Nested Schema for `spec.platform_properties.aws.aws_role_mappings.project_role_ref`
 
-Required:
-
-- `name` (String) The identifier of the meshProjectRole
```
grubmeshi, line 197: f: also double check if the backend requires setting this (should align with project_role_ref). I think for resources, optional/required is decided on the container. but I understand now that this is tricky here, and we need to introduce another flag in meshRefOptions that tells if the ref attribute is part of a set/container `InContainer` to support the documented "unknown at plan time" problem with a simple required = true. only if `InContainer` is true, the scheema makes the identifier optional but the description explains that this is validated at plan time to be required (additional description text).

docs/data-sources/platform.md
grubmeshi: Confirmed — yes, this is a genuine fix. This is the platform **data source**, where the whole spec (including `project_role_ref` inside `aws_role_mappings`) is read back from the backend, so it must be read-only. On `main` it was hand-rolled as **Required**, which is nonsensical for a backend-returned field; routing it through `meshRefByName(..., Output: true)` makes it Read-Only as it should be.

Now covered: `TestAccPlatformDataSource/02_aws` stands up an AWS platform and asserts `project_role_ref` (`{name, kind}`) inside `aws_role_mappings` on the data source. Green in the Go Acceptance Test run on this push.

docs/resources/platform.md
grubmeshi: Double-checked against the backend, and the change is right — but the real issue was the *container*, not the ref. `project_role_ref.name` is a required input on the backend (`MeshProjectRoleRef.name`, non-null). It only *looked* wrongly optional here because `aws_role_mappings` was modeled as a `List` while the backend stores AWS/GCP role mappings in a **map keyed by the meshProjectRole** (unordered, unique). I converted `aws`/`gcp` role mappings to `SetNestedAttribute` (+ retyped the client DTO to `types.Set[T]`); as a set element the identifier is correctly lenient (`Optional+Computed` via `InSet`, see the landingzone thread), and the description now states it is required and validated at plan time. This also removes a latent permanent no-op plan diff from map reordering. Acceptance test added asserting the resolved refs.

docs/resources/landingzone.md
grubmeshi: Done — implemented as an `InSet` flag on `meshRefOptions` (named for the actual constraint: only **sets** hash elements by whole value, so lists/standalone refs never needed the leniency). Backend confirms `project_role_ref.name` is required. The default (zero value) is now a plain `Required` identifier; refs that are set-nested or are themselves a set element opt in with `InSet: true`, which keeps the `Optional+Computed`+`AlsoRequires` form and appends a description note explaining it is required and validated during planning. Applied to all role-mapping refs (`aws`/`gcp` now that they are sets) and to `mandatory_building_block_refs`/`dependency_refs`; standalone refs (`integration_ref`, non-defaulted `runner_ref`) now render as `Required`.

docs/resources/building_block_definition.md
```diff
@@ -555,7 +564,7 @@ Read-Only:
 Optional:
 
 - `kind` (String) meshObject type, always `meshBuildingBlockDefinition`.
-- `uuid` (String) UUID of the meshBuildingBlockDefinition.
+- `uuid` (String) UUID (`metadata.uuid`) of `meshBuildingBlockDefinition`.
```
grubmeshi: f: i want to see some extra description here that this is actualy not optional and only so because of being `InSet`

.agents/skills/new-resource-datasource/SKILL.md
```diff
@@ -42,6 +42,22 @@ Mid-complexity, clean, and complete — prefer these over the large `building_bl
 6. **`internal/provider/<name>_resource_test.go`** — a `TestAcc<Name>` test (see below).
 7. `task generate` (docs) and update `CHANGELOG.md`.
 
+## meshObject reference attributes ({kind, uuid|name})
+
+Build any reference to another meshObject with **`meshRefByUuid` / `meshRefByName`**
+(`schema_utils.go`) — never hand-roll the `{kind, uuid|name}` block. Pass
+`meshRefOptions{Kind, Description}`:
+
+- zero value → **required input** (the common case for a resource's own spec refs);
```
grubmeshi: f: zero value (empty meshRefOptions) is not true, as Kind/Description is missing. also Below short description of InSet is missing (the godoc is more detailed).

docs/resources/building_block_definition.md
grubmeshi: Fixed in d481233 (fixed up into the meshRef commit). The `InSet` caveat now lives on the identifier (`uuid`/`name`) attribute in `schema_utils.go`, not the block description — set-element-typed refs (`dependency_refs`, `mandatory_building_block_refs`) are spread into a `NestedAttributeObject` that keeps only `.Attributes`, dropping the block description, so the block-level note never reached these docs. The identifier renders under "Optional:" in every case, so the note now shows:\n\n> UUID (`metadata.uuid`) of `meshBuildingBlockDefinition`. Required; optional here only so a computed reference can be used inside a set, and enforced at plan time.

.agents/skills/new-resource-datasource/SKILL.md
grubmeshi: Fixed in d481233. Reworked the meshRef section: `meshRefOptions{Kind, Description}` are now stated as always required (with why), the first bullet is reframed from "zero value" to "no flag", and an `InSet: true` bullet was added with a short description that points to the `meshRefOptions` godoc for the full rationale.

henryde (approved):
LGTM

PR #239, opened 2026-07-17
allow all dedicated assignment types for outputs on manual BB definition version inputs

j0g3sc:
Allow all dedicated (non-`NONE`) output `assignment_type`s — `SIGN_IN_URL`, `RESOURCE_URL`, `SUMMARY` in addition to `PLATFORM_TENANT_ID` — on manual building-block definition version outputs.

### Review feedback addressed
- `nonNoneOutputAssignmentTypes` is derived once from the full enum via a new `enum.Enum.Except` helper, and used for the schema description, `ValidateConfig`, and the tests. Adding an assignment type now flows in without editing each site.
- The manual-output "allowed" tests loop that derived set instead of one hardcoded case per type.
- The changelog entry moved from FIXES to FEATURES (it allows more, it does not fix a defect).

### Includes dependabot chore (closes #240)
This branch also carries the `actions/setup-go` 6.5.0 → 7.0.0 bump from #240 as a leading `chore(deps)` commit. Merging this brings the bump onto `main`, so Dependabot will auto-close #240 as superseded.

> Note on #240's red acceptance run: it failed only because **Dependabot-triggered workflow runs do not receive repository secrets** — `MESHSTACK_API_KEY`/`MESHSTACK_API_SECRET` were empty, so `DefaultTestPreCheck` failed for every acceptance test. It was not a code regression. On this normal branch the secrets are present and acceptance runs for real.


internal/provider/building_block_definition_resource.go
```diff
@@ -163,14 +163,16 @@ func (r *buildingBlockDefinitionResource) ValidateConfig(ctx context.Context, re
 	}
 
 	// Manual building blocks derive their outputs from the inputs on the backend (one output per input,
-	// assignment type NONE). The backend only honors outputs the user marks with assignment type
-	// PLATFORM_TENANT_ID; everything else it regenerates. So configuring any non-PLATFORM_TENANT_ID output
-	// for a manual building block is rejected here - omit it and let it be computed (see issues #131, #176).
+	// assignment type NONE). The backend regenerates every derived output as NONE except those the user
+	// marks with a non-NONE assignment_type (PLATFORM_TENANT_ID, SIGN_IN_URL, RESOURCE_URL, SUMMARY), which
```
grubmeshi: f: do not repeat the non-NONE values so often,

internal/provider/building_block_definition_resource_schema.go
```diff
@@ -426,7 +426,11 @@ func (r *buildingBlockDefinitionResource) Schema(_ context.Context, _ resource.S
 						MarkdownDescription: "Map of output definitions for the building block. Keys are output names, values are output configuration objects. " +
 							"Outputs define values that building blocks produce and can be consumed by other building blocks. " +
 							"If implementation type is " + client.MeshBuildingBlockImplementationTypeManual.Markdown() +
-							", outputs are computed from the API response, so omit this attribute entirely unless you want to specify a static `assignment_type = \"PLATFORM_TENANT_ID\"` as part of a landing zone.",
+							", outputs are computed from the API response, so omit this attribute entirely unless you want to mark how a derived output is used by giving it a dedicated `assignment_type` (one of " +
+							client.MeshBuildingBlockDefinitionOutputAssignmentTypePlatformTenantID.Markdown() + ", " +
```
grubmeshi: define an extra var `nonNoneOutputAssignmentTypes` derived from the full enum set (all except NONE) and use that throughout code, in particular here when building the description. ensure that adding an enum entry will not cause regression as that list is defined once

internal/provider/building_block_definition_resource_test.go
```diff
@@ -1224,19 +1224,31 @@ resource "meshstack_building_block_definition" "test" {
...
 			name:        "omitted assignment_type rejected for manual",
 			outputs:     `outputs = { tenant = { display_name = "Tenant", type = "STRING" } }`,
-			expectError: regexp.MustCompile(`may only assign PLATFORM_TENANT_ID`),
+			expectError: regexp.MustCompile(`must have a special assignment_type`),
 		},
 		{
 			name:    "PLATFORM_TENANT_ID output allowed for manual",
 			outputs: `outputs = { tenant = { display_name = "Tenant", type = "STRING", assignment_type = "PLATFORM_TENANT_ID" } }`,
 		},
+		{
```
grubmeshi: consider adding steps using a loop over new nonNoneOutputAssignmentTypes as well

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ FEATURES:
 
 FIXES:
 - `meshstack_building_block_definition`: `display_order` on `version_spec.inputs` and `version_spec.outputs` again counts towards a version's computed `content_hash`. Changing only `display_order` on a draft does not really need to rerun the building block, but hashing it keeps the backend simpler and the provider follows the same rule. The `content_hash` version was raised from 2 to 3 to go with this change. A `content_hash` written by an older provider is now recomputed at the current version instead of being treated as changed, so upgrading no longer shows false plan diffs or reruns building blocks whose definition is already released.
+- `meshstack_building_block_definition`: manual building blocks (`implementation.manual`) now accept any special output `assignment_type` — `SIGN_IN_URL`, `RESOURCE_URL`, and `SUMMARY` in addition to `PLATFORM_TENANT_ID` — to mark how a derived output is used. Previously only `PLATFORM_TENANT_ID` was allowed. The output key must match an input key; the backend still derives the output set from the inputs.
```
grubmeshi: i think that should be a feature allowing more instead of fixing sth

PR #242, opened 2026-07-20
feat: Improves error message for wrong manual outputs

tfelix:
Manual outputs are handled a bit special and are autogenerated depending on the inputs of a manual building block.
An empty output property is thus always wrong. This is now checked and handled with a good error message to let the user know what is wrong.


CHANGELOG.md
```diff
@@ -2,13 +2,10 @@
...
-- `meshstack_building_block` and `meshstack_building_block_v2`: `spec.building_block_definition_version_ref` now carries a `kind` field (computed, always `meshBuildingBlockDefinitionVersion`), consistent with every other meshObject reference. The `meshstack_building_block_definition` computed `version_latest`, `version_latest_release` and `versions` outputs, and the `meshstack_building_block_v2` / `meshstack_building_blocks` data sources, expose the matching `kind` too, so wiring a version output into a building block keeps working.
 
 FIXES:
+- `meshstack_building_block_definition`: a manual building block with an explicit empty `version_spec.outputs = {}` is now rejected at plan time with a clear message telling you to omit the attribute, instead of failing at apply with "provider produced inconsistent result after apply: new element ... has appeared". Manual building blocks derive one output per input on the backend, so an explicit empty map can never match the applied result; omitting `version_spec.outputs` lets it be computed ([#240](https://github.com/meshcloud/terraform-provider-meshstack/issues/240)).
```
grubmeshi: f: I've released v0.23.3 already, bump it to `.4` patch. sorry, i wasn't aware that you wanted to make this fix. for manual BBD outputs, we still allow setting a the output assignment type to sth not-NONE (sth which i've just extended to allow all non-None values). Is this also correctly working considering the "inconsistent result after apply"? at least there are acc tests on main which do work.

internal/provider/building_block_definition_resource.go
```diff
@@ -170,15 +163,30 @@ func (r *buildingBlockDefinitionResource) ValidateConfig(ctx context.Context, re
...
+	// empty map. Terraform forbids a provider from planning a computed value that diverges from a known
+	// configured value, so this surfaces at apply as "provider produced inconsistent result after apply:
+	// new element ... has appeared" (issue #240). Reject it here and tell the user to omit the attribute.
+	if len(outputs.Elements()) == 0 {
+		resp.Diagnostics.AddAttributeError(
+			outputsPath,
+			"manual building block outputs must be omitted, not an empty map",
+			"Manual building block definitions derive their outputs from the inputs automatically (one output "+
+				"per input). An explicit empty `outputs = {}` conflicts with those derived outputs and fails at "+
+				"apply with \"provider produced inconsistent result after apply\". Remove the version_spec.outputs "+
```
grubmeshi, line 191: d: considering the little special case that output assignment types can be configured for manual BBDs (see other comment)

grubmeshi (changes requested):
Thanks for this improvement, see me comments though.

internal/provider/building_block_definition_resource_test.go
```diff
@@ -1217,36 +1216,35 @@ resource "meshstack_building_block_definition" "test" {
...
 			outputs:     `outputs = { tenant = { display_name = "Tenant", type = "STRING" } }`,
-			expectError: regexp.MustCompile(`must have a special assignment_type`),
+			expectError: regexp.MustCompile(`may only assign PLATFORM_TENANT_ID`),
+		},
+		{
+			name:    "PLATFORM_TENANT_ID output allowed for manual",
+			outputs: `outputs = { tenant = { display_name = "Tenant", type = "STRING", assignment_type = "PLATFORM_TENANT_ID" } }`,
+		},
+		{
+			// Regression test for issue #240: an explicit empty map is not the same as omitting outputs.
```
grubmeshi: f: avoid referencing issues in code comments, be succinct on comments in general (i think no comment is fine here)

CHANGELOG.md
tfelix: Yes I mean technically it is fine to have 0 outputs or less outputs than inputs. 0 outputs probably never makes sense and is just now with what we need to work with the Terraform API. So there is still the option to specify only outputs A and B even though you have inputs A, B, C.

internal/provider/building_block_definition_resource.go
tfelix: I am not fully understanding you comment. The outputs are generated by the server backend and you either need to match them here (then you can also modify the assignment type as you suggest) or you can leave them fully out then you just use the ones coming from the server.

CHANGELOG.md
tfelix: Bumped to v0.23.4.

internal/provider/building_block_definition_resource_test.go
tfelix: Agreed, dropped the comment — the test name already says it. Also removed the same issue-reference from the `ValidateConfig` comment for consistency.

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
-# v0.23.3
+# v0.23.4
```
grubmeshi: f: this is wrong, add new changelog entry

grubmeshi (changes requested):
again, changelog is not correctly fixed (see also skill in this repo)

CHANGELOG.md
tfelix: okay sorry messed up the numbering.

grubmeshi (approved):

PR #251, opened 2026-07-22
refactor: derive List query params from struct json tags in WithUrlQuery

grubmeshi:
`internal.WithUrlQuery` now takes a single value instead of a `(key, value)` pair:

- a **filter struct passed by value** whose `json` tags name the query params — zero-value fields are dropped (an implicit `omitempty`), so an unset filter needs neither a pointer nor an `omitempty` tag, and a zero-value struct adds no params; or
- a **`map[string]string` / `map[string]any`** taken verbatim (every entry sent, including deliberate zeros like `page=0`), round-tripping unchanged.

All `List` clients that filter by a query/filter struct now hand that struct to `WithUrlQuery` by value; the query structs carry `json` tags naming their params. Behaviour-preserving: the produced query params are identical to before.

Base PR for the plural platform/landing-zone data sources stack (#244 stacks on this).

### DIFF
--- a/client/building_block_definition.go
+++ b/client/building_block_definition.go
@@ -76,13 +76,19 @@ func newBuildingBlockDefinitionClient(ctx context.Context, httpClient internal.H
 	}
 }
 
+type meshBuildingBlockDefinitionListQuery struct {
+	// IncludeAllPublished is always true here: list definitions published across the platform in
+	// addition to the workspace's own. (A false bool would be dropped by WithUrlQuery, which is fine —
+	// this endpoint is only ever called with it set.)
+	IncludeAllPublished bool    `json:"includeAllPublished"`
+	OwnedByWorkspace    *string `json:"ownedByWorkspace"`
+}
+
 func (c meshBuildingBlockDefinitionClient) List(ctx context.Context, workspaceIdentifier *string) ([]MeshBuildingBlockDefinition, error) {
-	var options []internal.RequestOption
-	options = append(options, internal.WithUrlQuery("includeAllPublished", "true"))
-	if workspaceIdentifier != nil {
-		options = append(options, internal.WithUrlQuery("ownedByWorkspace", *workspaceIdentifier))
-	}
-	return c.meshObject.List(ctx, options...)
+	return c.meshObject.List(ctx, internal.WithUrlQuery(meshBuildingBlockDefinitionListQuery{
+		IncludeAllPublished: true,
+		OwnedByWorkspace:    workspaceIdentifier,
+	}))
 }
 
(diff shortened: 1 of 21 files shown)


client/platform_type.go
```diff
@@ -76,12 +76,12 @@ func (c meshPlatformTypeClient) Delete(ctx context.Context, name string) error {
 }
 
 func (c meshPlatformTypeClient) List(ctx context.Context, category *string, lifecycleStatus *string) ([]MeshPlatformType, error) {
```
grubmeshi, line 83: f: introduce struct with two fields instead of building map internally

client/building_block_definition.go
```diff
@@ -77,12 +77,11 @@ func newBuildingBlockDefinitionClient(ctx context.Context, httpClient internal.H
 }
 
 func (c meshBuildingBlockDefinitionClient) List(ctx context.Context, workspaceIdentifier *string) ([]MeshBuildingBlockDefinition, error) {
```
grubmeshi, line 87: f: introduce struct with one field instead

client/project.go
```diff
@@ -60,13 +60,11 @@ func (c meshProjectClient) Read(ctx context.Context, workspace string, name stri
 }
 
 func (c meshProjectClient) List(ctx context.Context, workspaceIdentifier string, paymentMethodIdentifier *string) ([]MeshProject, error) {
```
grubmeshi, line 67: f: introduce struct with two fields, also see other places where this is missing and a map is constructed while a struct is a much better approach.

client/platform_type.go
grubmeshi: Done — introduced `meshPlatformTypeListQuery{ Category, LifecycleStatus *string }` and pass it to `WithUrlQuery` instead of building a map.

client/building_block_definition.go
grubmeshi: Done — introduced `meshBuildingBlockDefinitionListQuery`. It has two fields: `includeAllPublished` (a bool, always `true` here — it must be emitted, and a non-zero bool is sent) plus `ownedByWorkspace *string` (nil is dropped). Passed by value to `WithUrlQuery`.

client/project.go
grubmeshi: Done — introduced `meshProjectListQuery{ WorkspaceIdentifier string; PaymentIdentifier *string }`. Also swept the other internal map builders to structs: `platform_type`, `building_block_definition`, and `building_block_definition_version`. The one remaining map is the paginator's `page` (`mesh_object_client.go`) — kept on purpose because `page=0` must be sent verbatim and a struct would omit the zero value (this is the documented rationale for the map path in `WithUrlQuery`). Also added a note to AGENTS.md codifying the struct-for-query-params convention.

PR #253, opened 2026-07-23
feat: add non_ephemeral_secret provider function

grubmeshi:
## non_ephemeral_secret provider function

Adds a provider-defined function `provider::meshstack::non_ephemeral_secret(secret_value)` that builds the `sensitive` secret block used across the provider (`meshstack_platform`, `meshstack_integration`, `meshstack_building_block`, …) from a value kept in Terraform **config/state** rather than supplied via an `ephemeral` resource.

It sets:
- `secret_value` → the given value
- `secret_version` → `sha256(secret_value)` (hex)

So a change to the value changes the version and re-applies the write-only `secret_value` to the backend, while an unchanged value produces no diff. This packages the previously documented `secret_version = nonsensitive(sha256(<secret_value>))` workaround into a single call:

```hcl
access_token = provider::meshstack::non_ephemeral_secret(var.access_token)
```

For truly ephemeral secrets, keep setting `secret_value` from an `ephemeral` resource and manage `secret_version` yourself instead.

### Changes
- New function `internal/provider/non_ephemeral_secret_function.go`, registered in `provider.go`.
- Adapted the `meshstack_platform` AKS example (`resource_05_aks.tf`) — the very example that previously spelled out the `nonsensitive(sha256(...))` workaround — to use the function. Its existing acceptance test pins the exact `secret_version` hash and still passes.
- Added the function's own example (`examples/functions/non_ephemeral_secret/function.tf`) and regenerated docs (`task generate`).
- Test `TestAccNonEphemeralSecretFunction` (runs in mock/unit mode too) asserts the returned object across several inputs.
- CHANGELOG entry under the pending **v0.24.1**.

### Verification
- `go test ./...` (unit/mock) green; `golangci-lint run ./...` clean.
- Verified end-to-end with a dev-override build: the returned object assigns cleanly to a real `sensitive` attribute (computed `secret_hash` treated as optional), and `secret_version` matches `sha256sum`.


docs/functions/non_ephemeral_secret.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Wrap the argument in `nonsensitive(...)`** when it is a sensitive value (e.g. a `sensitive` variable): `access_token = provider::meshstack::non_ephemeral_secret(nonsensitive(var.access_token))`. This looks backwards — you are un-marking a secret — but it is correct and usually what you want. Terraform and OpenTofu conservatively mark the *entire* result of any function call sensitive as soon as one argument is sensitive, so without the wrap `secret_version` shows as `(sensitive value)` in plans and you lose this function's whole point: a legible, hash-based rotation trigger. (This is core-language behavior — the provider never sees the mark, and the plugin-framework function API has no per-return sensitivity to override it with.) `nonsensitive(...)` strips the mark only for this expression; the returned `secret_value` is immediately re-protected by the resource's own sensitive, write-only schema attribute (so it never lands in state), while the non-reversible `sha256` in `secret_version` stays visible.
```
henryde: > `nonsensitive(...)` strips the mark only for this expression; the returned `secret_value` is immediately re-protected by the resource's own sensitive, write-only

This sounds wrong. The secret_value is never marked as non-sensitive in the first place.

docs/functions/non_ephemeral_secret.md
henryde, line 1: I would structure this part like this:
- What the function does
- Why this usage of nonsentive is okay
- What to do better (this needs to be condensed).
- Example (without re-stating everything about ephemeral secrets etc.)

henryde (approved):

internal/provider/non_ephemeral_secret_function.go
```diff
@@ -0,0 +1,84 @@
...
+			"variable): `access_token = provider::meshstack::non_ephemeral_secret(nonsensitive(var.access_token))`. " +
+			"This looks backwards — you are un-marking a secret — but it is correct and usually what you want. " +
+			"Terraform and OpenTofu conservatively mark the *entire* result of any function call sensitive as soon " +
+			"as one argument is sensitive, so without the wrap `secret_version` shows as `(sensitive value)` in " +
+			"plans and you lose this function's whole point: a legible, hash-based rotation trigger. (This is " +
+			"core-language behavior — the provider never sees the mark, and the plugin-framework function API has " +
+			"no per-return sensitivity to override it with.) `nonsensitive(...)` strips the mark only for this " +
+			"expression; the returned `secret_value` is immediately re-protected by the resource's own sensitive, " +
+			"write-only schema attribute (so it never lands in state), while the non-reversible `sha256` in " +
+			"`secret_version` stays visible.",
```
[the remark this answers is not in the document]
grubmeshi: Good catch. I dropped the word non-reversible and reworded it: secret_version is now described as a sha256 of the value that you should publish only when revealing that hash is acceptable for your case, since a low entropy value can be guessed from its hash. The safety is now grounded in write-only rather than in the hash being irreversible.

docs/functions/non_ephemeral_secret.md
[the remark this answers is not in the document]
grubmeshi: Fixed at the source (the MarkdownDescription in non_ephemeral_secret_function.go) and regenerated with task generate, so this generated page no longer says non-reversible.

examples/functions/non_ephemeral_secret/function.tf
```diff
@@ -0,0 +1,34 @@
...
+        base_url     = "https://dev.azure.com"
+        organization = "my-organization"
+
+        # Wrapping the secret in nonsensitive() looks wrong, but it is correct here.
+        # Any function call with a sensitive argument makes Terraform/OpenTofu mark the WHOLE
+        # result sensitive, so secret_version would show as "(sensitive value)" in plans and you
+        # would lose the whole point of this function: a legible, hash-based rotation trigger.
+        # nonsensitive() strips the mark only for this expression; the returned secret_value is
+        # immediately re-protected by this attribute's own sensitive + write-only schema (so it
+        # never lands in state), while the non-reversible sha256 in secret_version stays visible.
```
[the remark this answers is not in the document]
grubmeshi: Reworded this comment too. It no longer calls the hash non-reversible and instead explains the write-only reasoning.

docs/functions/non_ephemeral_secret.md
grubmeshi: You are right, that wording was wrong: the value is never re-marked. Reworded it so the safety comes from secret_value being a write-only attribute (never stored in state, shown only as a placeholder in plans), and removed the re-protected phrasing.

docs/functions/non_ephemeral_secret.md
grubmeshi: Restructured along your outline: what the function does, then why the nonsensitive() usage is okay, then one line on what to do better (no secret at all via keyless auth, otherwise an ephemeral resource), then the example. Condensed the ephemeral and WIF guidance to a single sentence.

PR #255, opened 2026-07-23
feat: expiry date on workspace bindings

vzaslavsky:
Workspace group and user bindings now accept the expiry date.

I'll probably keep this open until we have actually rolled out meshStack 2026.30.0.


docs/resources/workspace_group_binding.md
```diff
@@ -42,6 +44,10 @@ resource "meshstack_workspace_group_binding" "example" {
 - `subject` (Attributes) Selects the group for this binding. (see [below for nested schema](#nestedatt--subject))
 - `target_ref` (Attributes) Selects the workspace to which this binding applies. (see [below for nested schema](#nestedatt--target_ref))
 
+### Optional
+
+- `expiry_date` (String) Optional expiry date for this binding as an ISO 8601 date (`YYYY-MM-DD`). After this date the binding is no longer effective.
```
grubmeshi: d: note that if omitted expiry never happens?

examples/resources/meshstack_workspace_user_binding/resource.tf
```diff
@@ -14,4 +14,6 @@ resource "meshstack_workspace_user_binding" "example" {
   subject = {
     name = "user@meshcloud.io"
   }
+
+  expiry_date = "2026-12-31"
```
grubmeshi, line 20: f: add little comment that this is optional and if omitted, it never expires (or some max default expiry is taken?)

internal/provider/workspace_group_binding_resource.go
```diff
@@ -98,6 +98,12 @@ func (r *workspaceGroupBindingResource) Schema(_ context.Context, _ resource.Sch
 					},
 				},
 			},
+
+			"expiry_date": schema.StringAttribute{
```
grubmeshi, line 102: d: could the backend also return some max expiry if that is configured for some mehstack instances? then this would become a computed attribute.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# v0.24.2
+
+Requires meshStack 2026.30.0 or later (previously 2026.29.0).
```
grubmeshi, line 3: n: I'm going to merge #244 soon which already bumps the min stack version.

grubmeshi (approved):
LGTM, see comments (minor things).

CHANGELOG.md
vzaslavsky: Ah, good to know. Then I'm going to rebase on top of that and will feel free to merge it in, too.

internal/provider/workspace_group_binding_resource.go
vzaslavsky: Absolutely correct -- some environments have recertification enabled for certain roles, which means a forced expiry date. I'll add all of that to the docs and comments.

PR #258, opened 2026-07-24
Feature/bd 2607 meshobject api ignores quotas

tfelix:


client/tenant_v4.go
```diff
@@ -159,10 +187,13 @@ type MeshTenantCreateMetadata struct {
...
-	LandingZoneRef   *NamedRef                  `json:"landingZoneRef" tfsdk:"landing_zone_ref"`
-	PlatformTenantId *string                    `json:"platformTenantId" tfsdk:"platform_tenant_id"`
-	Quotas           types.Set[MeshTenantQuota] `json:"quotas" tfsdk:"quotas"`
+	PlatformRef      UuidRef   `json:"platformRef" tfsdk:"platform_ref"`
+	LandingZoneRef   *NamedRef `json:"landingZoneRef" tfsdk:"landing_zone_ref"`
+	PlatformTenantId *string   `json:"platformTenantId" tfsdk:"platform_tenant_id"`
+	// RequestedQuotas is the preferred key->value form; Quotas is the deprecated list form. Only one
+	// should be set — the backend rejects a create that carries both with conflicting values.
+	RequestedQuotas map[string]RequestQuotaValue `json:"requestedQuotas,omitempty" tfsdk:"requested_quotas"`
+	Quotas          types.Set[MeshTenantQuota]   `json:"quotas,omitempty" tfsdk:"quotas"`
```
sttomm: d: Can we mark that more obvious as deprecated? AT least with a dedicated comment right here on that field?

client/tenant_v4.go
```diff
@@ -129,25 +134,48 @@ type MeshTenantMetadata struct {
...
 	PlatformWorkspaceId    *string             `json:"platformWorkspaceId" tfsdk:"platform_workspace_id"`
 	Tags                   map[string][]string `json:"tags" tfsdk:"tags"`
+	// AppliedQuotas are the effective quotas meshStack applied to the tenant as a key->value map, each
+	// value a structured object (e.g. `{"limits.cpu": {"value": 4}}`). spec.requested_quotas carries
+	// only the values requested at create (create-only); the effective quotas here can differ once
+	// landing-zone defaults are merged in or an operator adjusts them, so drift is tracked against these.
+	AppliedQuotas map[string]AppliedQuotaValue `json:"appliedQuotas" tfsdk:"applied_quotas"`
 }
 
 type MeshTenantQuota struct {
```
sttomm, line 163: d: This is the deprecated quota type for the deprecated Quota property?

client/tenant_v4.go
```diff
@@ -129,25 +134,48 @@ type MeshTenantMetadata struct {
...
+// change to the requested_quotas map shape.
+type RequestQuotaValue struct {
+	Value int64 `json:"value" tfsdk:"value"`
+}
+
+// AppliedQuotaValue is a tenant quota value as actually applied by the backend. Kept distinct from
+// RequestQuotaValue so it can later carry applied-only context (e.g. why the applied value differs
+// from what was requested).
+type AppliedQuotaValue struct {
+	Value int64 `json:"value" tfsdk:"value"`
```
sttomm, line 182: d: Do we really need RequestQuotaValue and AppliedQuotaValue as separate structs? Could be just one QuotaValue as they are identical, right?

docs/data-sources/tenant.md
```diff
@@ -88,13 +89,29 @@ Read-Only:
...
+- `value` (Number)
+
+
 
 <a id="nestedatt--status"></a>
 ### Nested Schema for `status`
 
 Read-Only:
 
+- `applied_quotas` (Attributes Map) The effective quotas meshStack applied to this tenant, as a map keyed by quota key whose value is an object carrying the applied `value`. (see [below for nested schema](#nestedatt--status--applied_quotas))
```
sttomm: d: We should describe that requested quota may not be applied righta way as an operator may need to approve them first

internal/clientmock/mock_tenant_v4.go
```diff
@@ -33,6 +33,13 @@ func (m MeshTenantV4Client) Create(_ context.Context, tenant *client.MeshTenantV
 	// Simulate a successful tenant creation with platformTenantId set
 	tenantName := tenant.Metadata.OwnedByWorkspace + "." + tenant.Metadata.OwnedByProject + "." + tenant.Spec.PlatformIdentifier
 
+	// The mock applies requested quotas verbatim, so effective status.appliedQuotas mirrors spec.quotas.
```
sttomm, line 36: d: WOuld it make sense to also have a test that actually tests it works when tehy are different?

internal/provider/tenant_model.go
```diff
@@ -38,13 +38,14 @@ type tenantResourceModel struct {
...
-// backend defaults or reorders quotas.
-func tenantResourceModelFromDto(dto *client.MeshTenant, specQuotas clientTypes.Set[client.MeshTenantQuota], waitForCompletion bool) tenantResourceModel {
+// tenantResourceModelFromDto builds the resource state from an API response. specQuotas and
+// specRequestedQuotas are the known (configured) spec quota fields carried from plan/state: both are
+// Optional (not computed), so they must echo the configured value verbatim to avoid an
+// inconsistent-result-after-apply error when the backend defaults or reorders quotas.
+func tenantResourceModelFromDto(dto *client.MeshTenant, specQuotas clientTypes.Set[client.MeshTenantQuota], specRequestedQuotas map[string]client.RequestQuotaValue, waitForCompletion bool) tenantResourceModel {
 	spec := dto.Spec
-	spec.Quotas = specQuotas
+	spec.Quotas = specQuotas //nolint:staticcheck // bridging the deprecated quotas field for backward compatibility
```
sttomm, line 47: d: But iof we still provide those and they are determiend by the backend in the response from teh applied quotas, tf will still fail as it changed?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
...
...ason. The old `spec.quotas` set of `{key, value}` objects is **deprecated** but still accepted for backward compatibility; provide either `requested_quotas` or `quotas`, not both with conflicting values (the backend rejects a mismatch, or a `quotas` list carrying duplicate keys, with HTTP 400). `spec.requested_quotas` remains create-only — the meshTenant API has no quota update endpoint, so changing it on an existing tenant is still rejected at plan time; use it only to request quotas at creation. It is also a create-time input the API does not return on read, so the data sources expose it as typically null; read effective values from `status.applied_quotas`. meshStack now enforces each quota's `[minValue, maxValue]` bounds (for every caller) and the auto-approval threshold (for non-admin API keys) at creation, rejecting out-of-range or above-threshold requests with a descriptive HTTP 400 that the provider surfaces. Requires the backend carrying these changes (meshStack 2026.30.0). The deprecated `meshstack_tenant_v4` resource and data source now also expose `status.applied_quotas` (a map of `{ value }` objects) read back from the API instead of echoing the requested `spec.quotas`.
```
sttomm: d: wow, that's a mouthful of text. I did not read it ... Can it probably be shortened?

sttomm (approved):
I just had a top level look as the predecessor PR was already reviewed. And in general the separation of RequestedQuota and AppliedQuota sounds reasonable to me

client/tenant_v4.go
tfelix: Yes I can add a comment.

client/tenant_v4.go
tfelix: Not sure if this is a good idea for the TF provider but I will ask AI.

docs/data-sources/tenant.md
tfelix: Yeah but I think this is not yet full implemented for the meshObject API. I need to cross check with the meshObject API how this behaves. I think for now if it does not fit into the auto-allow window it is denied. But the idea is to have this request cycle managed via the meshObject API in the end.

client/tenant_v4.go
tfelix: Done — `MeshTenantCreateSpec.Quotas` now carries a proper `// Deprecated:` godoc marker, the same as `MeshTenantSpec.Quotas` already had, so IDEs strike it through and staticcheck flags it.

That surfaced two more bridging use sites in the tenant mock, which now carry an explicit `//nolint:staticcheck // the mock must keep serving the deprecated quotas form`.

client/tenant_v4.go
tfelix: Yes. `MeshTenantQuota` is the `{key, value}` element of the list-form `spec.quotas` and has no other user — the deprecated field on `meshstack_tenant` plus the quotas of the deprecated `meshstack_tenant_v4`. Added a doc comment on the type saying exactly that.

I deliberately did *not* put a `// Deprecated:` marker on the type itself: it is still the live quota shape of `meshstack_tenant_v4`, so marking it would mean a `//nolint:staticcheck` at every use site that intentionally keeps that resource working. The comment now states that too, so the omission does not read as an oversight.

client/tenant_v4.go
tfelix: You're right that they're identical — deliberately so, and I've kept them split but replaced the rationale, because the one in the comment was weak (speculative "might carry more fields later").

The real reason is compile-time: the resource has to echo the *configured* request in `spec` while reading effective values from `status`. With a single shared `QuotaValue`, `spec.RequestedQuotas = dto.Status.AppliedQuotas` compiles happily — and that assignment is exactly the requested-vs-effective conflation this PR exists to fix. Separate types turn it into a build error. The comment now says that instead.

Merging would be about 8 lines cheaper (the two DTOs, the mock's conversion, one function signature) and would not touch the Terraform schema or JSON, so it stays reversible either way. If you'd rather have the smaller surface, say so and I'll collapse them.

docs/data-sources/tenant.md
tfelix: I cross-checked this against the backend, and it turns out the opposite is true for the meshObject API — so instead of adding this, I fixed the text we already had, which was making your point and was wrong.

`MeshTenantService.validateQuotaValuesOrThrow` documents it explicitly:

> Above-threshold requests are rejected here rather than queued for operator approval as the Panel does. This is deliberate: the meshObject API has no quota-request representation, and a pending request would let a Terraform apply report success while the requested quotas are not in effect.

Concretely, on create: unknown keys and values outside `[minValue, maxValue]` are HTTP 400 for every caller; an increase beyond the platform's auto-approval threshold is HTTP 400 too, unless the principal has `ADM_TENANT_SAVE` (measured against the tenant's current quotas, or the landing zone's defaults for a new tenant, so reductions always pass). Anything accepted is applied verbatim — `resolveEffectiveQuotas` = landing-zone defaults + requested overrides. There is never a pending quota request via this API.

So a requested key can only diverge on a *later* refresh, after an operator changed the tenant or a panel quota request was approved. Our warning claimed "needs a platform operator to approve the quota request before it takes effect", which would have sent people looking for a pending request that cannot exist — rewritten. The `spec.requested_quotas` and `status.applied_quotas` schema descriptions (hence the generated docs you commented on) now explain that applied is a superset of requested via landing-zone defaults, that a differing requested key means a post-creation change, and that out-of-bounds/above-threshold requests are refused rather than queued. Fixed the same claim in the CHANGELOG.

internal/clientmock/mock_tenant_v4.go
tfelix: Good catch, and agreed this belongs in a real scenario test against the backend rather than a mock trick. Checking what the backend can actually be made to produce, two of the three cases are reachable live and one isn't:

- **applied ⊋ requested** — new `quotas_landing_zone_defaults` subtest. The landing zone defines a `limits.memory` default that the tenant never requests, so `status.applied_quotas` genuinely carries more than `spec.requested_quotas` (`resolveEffectiveQuotas` = LZ defaults + requested overrides). It runs in **both** modes: the tenant mock now overlays the assigned landing zone's default quotas the same way the backend does, so unit mode exercises the same shape.
- **above the auto-approval threshold** — new `quotas_above_auto_approval_threshold` subtest (acceptance-only, like `quotas_out_of_range`). This guardrail had no coverage at all.
- **a different *value* for a requested key** is unreachable through the meshObject API by design (see the `docs/data-sources/tenant.md` thread): accepted requests are applied verbatim, and only an operator action outside the API can change them afterwards. No provider client call can set that up, so it stays covered by `TestQuotaRealizationWarning`'s `value differs` / `not yet applied` cases.

One limitation worth noting: terraform-plugin-testing has no warning assertion, so the end-to-end tests check state plus an empty follow-up plan; the warning text itself is asserted in the unit test.

internal/provider/tenant_model.go
tfelix: No, and this line is precisely what prevents it. `spec.requested_quotas` (and the deprecated `spec.quotas`) are `Optional` but **not** `Computed`, so they must come back exactly as configured — that is what these two assignments do, echoing the value carried in plan/state instead of whatever the backend returned in its spec. The backend-determined values live in the separate `Computed` `status.applied_quotas`, so they never flow into `spec` and cannot produce a diff. Dropping the echo is what would give "Provider produced inconsistent result after apply".

The new `quotas_landing_zone_defaults` test pins this down: step 1 creates a tenant requesting `limits.cpu` on a landing zone that defaults `limits.memory`, step 2 is a `PlanOnly` re-plan after refresh, asserting `spec.requested_quotas` still holds only the configured `limits.cpu`, `status.applied_quotas` holds both keys, and the plan is empty — no drift.

CHANGELOG.md
tfelix: Fair — it was one ~2,000-character paragraph. Cut to roughly a quarter, in two bullets: one for the map-shaped quotas (`spec.requested_quotas` create-only, `status.applied_quotas` incl. landing-zone defaults, `spec.quotas` deprecated, the 2026.30.0 guardrails), one for `meshstack_tenant_v4`. The mechanics that were spelled out here now live in the schema docs, where people actually look them up.

It also carried the incorrect "a pending platform-operator approval may be the reason" claim, which is gone — see the `docs/data-sources/tenant.md` thread for why.

PR #260, opened 2026-07-28
Feature/workspace tag management

florianow:


j0g3sc (approved):
I made "some" changes, but it all boils down to additional tests and documentation.

Listed a few concerns [here](https://github.com/meshcloud/terraform-provider-meshstack/pull/260#issuecomment-5131469083), but I think the new resources are valuable nonetheless if you know what you do.

PR #266, opened 2026-08-07
feat: add meshstack_instance data source

### LINKED ISSUE #221: feat: data source for meshStack instance metadata (endpoint, version)
Canonical feature request (see there for the full motivation, not repeated here):
https://feedback.meshcloud.io/feature-requests/p/terraform-provider-data-source-for-meshstack-instance-metadata

## Concrete pain point from our own e2e/smoke-test setup

The provider is already configured with the meshStack endpoint, but Terraform config cannot read it back. Any module that needs the endpoint *value* — e.g. the [meshstack-hub noop runner backplane](https://github.com/meshcloud/meshstack-hub/blob/main/modules/meshstack/noop/backplane/variables.tf), which templates it into a `runner-config.yml` as the runner's API polling URL — must declare its own `meshstack_endpoint` input variable.

That single variable then has to be threaded through every harness that invokes such a module. In our (internal) smoke-test repository we currently export `TF_VAR_meshstack_endpoint` in two places — the CI workflow env and the local developer setup script — purely to feed a value the provider already knows. Each carries a comment pointing at the Canny ticket above.

## What a data source would simplify

With something like

```hcl
data "meshstack_instance" "this" {}

# data.meshstack_instance.this.endpoint
# data.meshstack_instance.this.version   (as exposed by /mesh/info)
```

modules could read the endpoint from the provider configuration directly:

- the hub runner backplane drops its `meshstack_endpoint` variable,
- both `TF_VAR_meshstack_endpoint` exports in the smoke-test harness (CI + local dev script) get deleted,
- e2e tests could additionally assert against the instance version/build metadata where relevant.

## Backing endpoint & attribute schema

The data source reads meshfed's `GET /mesh/info` (public, unauthenticated — works fine with the provider's existing API-key client). All attributes are computed/read-only:

```hcl
data "meshstack_instance" "this" {}
```

| Attribute | Type | Source |
|---|---|---|
| `endpoint` | `string` | provider config |
| `version` | `string` | `/mesh/info` |
| `is_four_eyes_enabled` | `bool` | `/mesh/info` (`is4EPEnabled`) |
| `metadata` | `map(string)` | `/mesh/info` — meshComponent → git commit SHA (subsystem revs; drives runner image tags) |
| `admin_workspace_identifier` | `string` | **new** — needs adding to `/mesh/info` |

`admin_workspace_identifier` is the admin/partner workspace (the first workspace created; from backend config `web.register.default-partner-identifier`) — the only valid owner for admin-only resources like Entra ID integrations, and currently undiscoverable via any API. Exposing it lets configs (and our `04_entra_id` acceptance test, which hardcodes it today) stop hardcoding the workspace name. Note `/mesh/info` is public, so this would be exposed unauthenticated — acceptable if treated as low-sensitivity, otherwise gate it behind auth.

j0g3sc:
## Summary

Implements #221: a new `meshstack_instance` data source exposes information about the meshStack instance the provider is configured against, so modules can read it directly instead of threading a separate `meshstack_endpoint` variable through every caller.

```hcl
data "meshstack_instance" "this" {}
```

- `endpoint` — from the provider configuration
- `version`, `is_four_eyes_enabled`, `metadata` (subsystem git commit SHAs) — from the public, unauthenticated `/mesh/info` endpoint
- `admin_workspace_identifier` — also from `/mesh/info` (the admin/partner workspace, the only valid owner for admin-only resources such as Entra ID integrations)

## Implementation notes

- New `client.MeshInfoClient` (`client/mesh_info.go`) wraps `/mesh/info`; it shares a `fetchMeshInfo` helper with the existing startup version check in `client/client.go`, which previously declared its own inline, narrower response type.
- `admin_workspace_identifier` assumes the backend's `/mesh/info` response includes `adminWorkspaceIdentifier` — this field does not exist on `/mesh/info` yet as of the current `meshfed-release` `MeshInfoController`. A companion backend change adding it is expected to land separately; per this repo's conventions for preview/in-flight backend APIs, that PR should use a matching `feature/`-prefixed branch name so CI pairs the two. Until that lands, this attribute reads back as an empty string against a real backend.

## Test plan

- [x] `go build ./...`
- [x] `task lint` (0 issues)
- [x] `TestAccInstanceDataSource` passes in mock mode (`go test ./internal/provider/...`)
- [ ] Acceptance run against a real backend (`task testacc`) will fail on `admin_workspace_identifier` until the companion `meshfed-release` change lands


client/mesh_info.go
```diff
@@ -0,0 +1,63 @@
...
+	AdminWorkspaceIdentifier string            `tfsdk:"admin_workspace_identifier"`
+}
+
+// meshInfoDto is the raw /mesh/info response shape.
+type meshInfoDto struct {
+	Version                  version.Version   `json:"version"`
+	Is4EPEnabled             bool              `json:"is4EPEnabled"`
+	Metadata                 map[string]string `json:"metadata"`
+	AdminWorkspaceIdentifier string            `json:"adminWorkspaceIdentifier"`
+}
```
grubmeshi, line 27: f: combine this into a single struct `MeshInfo`, using `tfsdk:"-"` to ignore things. amend missing fields in Read method below.
```suggestion
```

client/mesh_info.go
```diff
@@ -0,0 +1,63 @@
...
+	return &MeshInfo{
+		Endpoint:                 c.httpClient.RootUrl.String(),
+		Version:                  dto.Version.String(),
+		IsFourEyesEnabled:        dto.Is4EPEnabled,
+		Metadata:                 dto.Metadata,
+		AdminWorkspaceIdentifier: dto.AdminWorkspaceIdentifier,
+	}, nil
+}
+
+func fetchMeshInfo(ctx context.Context, httpClient internal.HttpClient) (meshInfoDto, error) {
```
grubmeshi, line 56: f: I think we can use Read method directly in client.go instead of private call to fetchMeshInfo (remove this private emthod and only use Read method using combined MeshInfo dto (see other comment)

the trick would be to construct meshInfoClient struct already as local var in client.go New and pass it to getting / checking the version, the also hand it out to client used by the provider data source.

internal/clientmock/mock_mesh_info.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// MockEndpoint is the fixed endpoint the mock client reports for meshstack_instance: the mock
+// client factory (see ApplyAndTest) never sees the provider's actual configured endpoint, since it
+// bypasses newProviderClient entirely.
+const MockEndpoint = "http://localhost:8080"
+
+const MockMeshVersion = "2026.30.0"
+
+const mockAdminWorkspaceIdentifier = "demo-partner"
```
grubmeshi, line 16: f: inline those constants into `Read` below? are they reused at all? if only for assertions, duplicate the expected value there.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v0.24.4
+
+FEATURES:
+- New `meshstack_instance` data source exposes information about the meshStack instance the provider is configured against: `endpoint` (from the provider configuration), and `version`, `is_four_eyes_enabled`, `metadata` (subsystem git commit SHAs) and `admin_workspace_identifier` (the admin/partner workspace) from the public, unauthenticated `/mesh/info` endpoint. Lets modules read the endpoint the provider is already configured with instead of threading a separate `meshstack_endpoint` variable through every caller, and resolves the admin workspace without hardcoding its identifier. `admin_workspace_identifier` requires a meshStack version that returns `adminWorkspaceIdentifier` from `/mesh/info`; it reads as an empty string on older instances.
```
grubmeshi, line 4: f: make this less verbose and refer to the added docs of the data source for specific attributes.

docs/data-sources/instance.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Read-Only
+
+- `admin_workspace_identifier` (String) Identifier of the admin (partner) workspace, the only valid owner for admin-only resources such as `meshstack_integration` Entra ID integrations.
+- `endpoint` (String) URL of the meshStack API this provider is configured against.
+- `is_four_eyes_enabled` (Boolean) Whether the four-eyes principle (role approval) is enabled on this meshStack instance.
```
grubmeshi, line 26: f: i'd like to expose this in the schema a bit more general as `enabled_feature_flags`, which entries such as `"four_eyes_role_approval"` as the only currently possible entry of this "set of strings" attribute.

grubmeshi (changes requested):
LGTM in general, see my comments though.

grubmeshi (approved):
LGTM, one little comment about the added "pr-review-comments" skill.

Issue #148, opened 2026-04-10
Add missing dependency data sources and reusable ref outputs


grubmeshi:
Context: To support data-source-first wiring in examples and real modules, several dependency lookups still require hardcoded identifiers or ad-hoc expressions.

Gaps:
- Missing data sources: meshstack_location, meshstack_workspace_group_binding, meshstack_workspace_user_binding.
- Missing singular/plural parity in a few domains.
- Missing reusable read-only ref-style outputs on some resources/data sources where cross-resource wiring currently needs metadata.uuid extraction.

Expected outcome:
- Add missing data sources where feasible.
- Add or standardize reusable computed outputs (ref and similar helper outputs) for cross-resource wiring.
- Update docs/examples to prefer these outputs over hardcoded identifiers.

Notes:
- Keep behavior backward compatible.
- Follow existing patterns like meshstack_workspace.ref, meshstack_tenant_v4.ref, and BBD version_latest / version_latest_release.

PR #273, opened 2026-08-10
feat!: reference parent building blocks by ref, add the building block ref output and a singular data source

grubmeshi:
## Summary

Three related changes to how building blocks are referenced and read, plus one commit that only
touches comments.

**1. A parent is a plain meshObject reference.** On `meshstack_building_block`,
`spec.parent_building_blocks` is renamed to `spec.parent_building_block_refs`, and each element
changes from `{buildingblock_uuid, definition_uuid}` to a `{kind, uuid}` ref:

```hcl
# before
parent_building_blocks = [{ buildingblock_uuid = "1111…", definition_uuid = "2222…" }]

# after
parent_building_block_refs = [meshstack_building_block.parent.ref]   # or [{ uuid = "1111…" }]
```

This is the element shape that `mandatory_building_block_refs` and `dependency_refs` already use.
meshStack derives the definition from the referenced block, so one ref replaces two attributes. The
new name says what the attribute holds and matches the meshBuildingBlock API field it fills,
`spec.parentBuildingBlockRefs`. It ships in the same release as the reshape, so an operator edits
each of these blocks once instead of twice.

**State migrates automatically**: the resource schema moves to version 1, and the v0 upgrader in
`internal/provider/building_block_upgrade.go` reads the old attribute out of prior state and writes
the refs into the new one. No replacement is planned for an existing building block. The upgrader
also fills the new `ref` output, because prior state has no value for it.

**2. `meshstack_building_block` gains a computed `ref` output**, which you can put directly into
another block's `spec.parent_building_block_refs`.

**3. A singular `meshstack_building_block` data source** that reads one block by UUID. The
`meshstack_buildingblock` and `meshstack_building_block_v2` data sources are **deprecated** in its
favour. Those two are the only way to read a block's `status.outputs` by UUID today, and
`meshstack_building_blocks` cannot filter by UUID, so they needed a replacement rather than a plain
removal. What remains is the new singular data source (by UUID) and the plural one (filtered lists).

Four commits, in that order, with a leading `docs:` one that records why importing a building block
cannot recover `content_hash`. The hash is a pure function of the referenced definition version's
spec, but reading that version needs access to the workspace that owns the definition, which the
importing credentials need not have. Recovery would therefore work for some API keys and not for
others.

## Why this is breaking, and why it ships now

This is the bottom of a stack. @sttomm's review on meshcloud/meshfed-release#10418 established that
the backend cannot drop its deprecated flat parent fields until a **released** provider can express
a parent as a ref. This PR is that release.

**It deliberately does not raise `MinMeshStackVersion`**, which stays `2026.30.0`. Its whole purpose
is to install on the backend that runs today. The resource sends the parents under **both** wire
field names, `parentBuildingBlockRefs` and the deprecated `parentBuildingBlocks`, and reads back
whichever of the two the backend serves:

| backend | write | read |
|---|---|---|
| without `parentBuildingBlockRefs` | reads the deprecated `parentBuildingBlocks`, ignores the field it does not know | `parentBuildingBlocks` |
| with `parentBuildingBlockRefs` | reads `parentBuildingBlockRefs`; the deprecated field names the same building blocks | `parentBuildingBlockRefs` |

Sending the unknown field is safe because the meshObject API does not reject unknown properties. Both
lists always name the same building blocks, so a backend that reads either one provisions the same
parents.

The compatibility code is confined to `MeshBuildingBlockV2Spec`'s `MarshalJSON` and `UnmarshalJSON`
in `client/building_block_v2.go`, so #274 can delete it in one place. Round-trip tests cover a
response that carries both fields and a response that carries only the deprecated one, and assert
that a request goes out under both names. The parent request format previously had **no** test
coverage at all, because the client mock never marshals JSON.

## Scope

`meshstack_building_block` and the `meshstack_building_blocks` data source move to
`parent_building_block_refs` and the ref shape. The deprecated `meshstack_building_block_v2` resource
and data source are **unaffected**: they keep their flat `spec.parent_building_blocks` with
`buildingblock_uuid` and `definition_uuid`, and they map to and from the shared client DTO
explicitly, so the deprecated attributes report the definition uuid as before. The
`meshstack_buildingblock` (v1) resource keeps its own flat shape; only its data source gains the
deprecation notice. #274 removes all of them.

## Stack

1. **this PR** → `v0.24.4`, released before the backend GA flip merges.
2. meshcloud/meshfed-release#10613 (v2-preview media type) and meshcloud/meshfed-release#10612 (GA
   media type) on the backend side.
3. #274 — provider GA media-type flip, `v0.26.0`, based on this branch.

Closes item 4 of #148, which assigned both halves of the building-block reference case here.


client/building_block_v2.go
```diff
@@ -53,8 +53,61 @@ type MeshBuildingBlockV2Spec struct {
 	DisplayName                       string                                  `json:"displayName" tfsdk:"display_name"`
 
 	// Inputs as pointer MeshBuildingBlockInput to support mocking secret responses.
-	Inputs               map[string]*MeshBuildingBlockInput `json:"inputs" tfsdk:"inputs"`
-	ParentBuildingBlocks types.Set[MeshBuildingBlockParent] `json:"parentBuildingBlocks" tfsdk:"parent_building_blocks"`
+	Inputs                  map[string]*MeshBuildingBlockInput   `json:"inputs" tfsdk:"inputs"`
+	ParentBuildingBlockRefs types.Set[MeshBuildingBlockV2Parent] `json:"parentBuildingBlocks" tfsdk:"parent_building_block_refs"`
```
grubmeshi: f: align this with backend (drop `_refs` throughout)

```suggestion
	ParentBuildingBlocks types.Set[MeshBuildingBlockV2Parent] `json:"parentBuildingBlocks" tfsdk:"parent_building_blocks"`
```

internal/provider/building_block_data_source.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/meshcloud/terraform-provider-meshstack/internal/types/generic"
+	"github.com/meshcloud/terraform-provider-meshstack/internal/types/secret"
+)
+
+var (
+	_ datasource.DataSource              = &buildingBlockDataSource{}
+	_ datasource.DataSourceWithConfigure = &buildingBlockDataSource{}
+)
+
+func NewBuildingBlockDataSource() datasource.DataSource {
```
grubmeshi, line 24: f: this datasource addition shall have it's own commit (PR should have two commits at least), see other comments though

examples/resources/meshstack_building_block/resource_01_workspace.tf
```diff
@@ -17,6 +17,10 @@ resource "meshstack_building_block" "example_workspace" {
         value = jsonencode("dev")
       }
     }
+
+    # Building blocks can depend on each other: a parent's outputs feed this block's inputs.
+    # Reference a parent by its computed `ref`.
+    # parent_building_block_refs = [meshstack_building_block.parent.ref]
```
grubmeshi: f: easy to miss when refactoring

```suggestion
    # parent_building_blocks = [meshstack_building_block.parent.ref]
```

internal/provider/acctest/testconfig/build_building_block.go
```diff
@@ -26,6 +26,40 @@ func BBWorkspace(t *testing.T) (config Config, buildingBlockAddr Traversal, buil
...
+	)
+
+	// Both blocks come from the same example, so each needs its own block label.
+	buildingBlock := func(name string, extra ...ExpressionConsumer) (buildingBlockConfig Config, buildingBlockAddr Traversal) {
+		return Resource{Name: "building_block", Suffix: "_01_workspace"}.Config(t).WithFirstBlock(
+			append([]ExpressionConsumer{
+				RenameKey(name),
+				ExtractAddress(&buildingBlockAddr),
+				Descend("spec", "display_name")(SetString("my-" + name + "-building-block")),
+				// Wire only the version uuid — see BBWorkspace.
```
grubmeshi, line 50: d: i'd like to challenge this workaround and actually fix the import to support `content_hash` calucation also on import, as on import the BBD with it's versions should be read in and the content_hash can be computed in tf provider, right? if so, implement this as another, separate commit as part of this PR (very first fix commit)

grubmeshi: Feasible on the API side — the backend does serve `GET /meshobjects/meshbuildingblockdefinitionversions/{uuid}`, and the hash is a pure function of that version's spec, so it would reproduce exactly (the definition resource already hashes a read-back version).

The blocker is permissions: reading a definition version requires access to the workspace owning the definition, which the credentials importing a building block need not have — a workspace-scoped key can consume a block whose definition lives elsewhere. Recovery would then succeed for some keys and not others, making imported state depend on who imported it. So the workaround stays and the catch is now recorded in a comment on `buildingBlockResource.ImportState`, as the first commit of this PR.

Worth noting for the original concern: with `content_hash` simply absent from the config, the post-import apply already triggers no run — `rerunNeeded` only compares hashes when the *plan* value is non-nil. The null in state only costs one rerun when the config does set it.

client/building_block_v2.go
grubmeshi: Done — `parent_building_blocks` throughout, so the attribute keeps the name the meshObject API uses and only its element shape changes. That also turns the state upgrader from a rename into a pure reshape, and makes the `v0.24.4` CHANGELOG entry a migration of the element rather than of the attribute. Scoped to the parent attribute: `target_ref`, `building_block_definition_version_ref` and `dependency_refs` do carry `Ref` in their backend names.

PR #283, opened 2026-08-18
feat: github platform type removed

vzaslavsky:
It's being merged into the custom type, as it was, essentially, treated like one.


grubmeshi (approved):
Although still draft, approving it now. Sounds legit.

PR #284, opened 2026-08-18
feat!: consume the meshStack API client from meshstack-cli

grubmeshi:
Moves `client/` out of this repository into
[meshcloud/meshstack-cli](https://github.com/meshcloud/meshstack-cli) and imports it from there, so
the provider and the meshStack CLI share one client instead of each carrying its own.

Companion PR: **meshcloud/meshstack-cli#1**, which scaffolds that repository and receives the client.

ClickUp: [86cb61rzz](https://app.clickup.com/t/86cb61rzz), milestone
[86cb61we4](https://app.clickup.com/t/86cb61we4).

## ⚠️ CI is red on purpose, and one commit is worth pushing on its own

`go.mod` carries a temporary `replace github.com/meshcloud/meshstack-cli => ../meshstack-cli`. That
path does not exist on a runner, so **every Go job here fails until the module is pinned**. The
directive is what allowed the tests to be run green locally before anything was pushed to the new
repository. Once meshstack-cli#1 merges: `go get github.com/meshcloud/meshstack-cli@main`, drop the
directive, commit `go.mod`/`go.sum`, and CI goes green.

**`fix: close the depguard patterns that silently governed nothing` (first commit) is independent of
all of that** and does not touch the client move. It is safe to review and push straight to `main`:

```shell
git push origin f8834af:main   # or the equivalent after any rebase
```

## Commit 1 — the depguard patterns governed nothing

Every directory pattern in the depguard rules used only the `**/dir/**/*.go` form. That matches files
in *subdirectories* of `dir` and never files directly inside it, so in a flat package it governs
nothing. Because a rule that matches no file reports no issue, this is indistinguishable from a rule
that passes.

Three sets of files were unchecked:

| Files | What was wrong |
|---|---|
| `client/*.go` (51 of the client's 60 files) | The `client` rule that keeps the client free of external dependencies only ever applied to `client/internal` and `client/types`. The top-level files fell through to the `provider` rule, which allows all of terraform-plugin-framework. |
| `client/*_test.go` | Same, under `client-test`. |
| `internal/clientmock/*.go` (all 23) | Matched **no rule at all**: `provider` excluded them with a single-star pattern that does match them, while `provider-test` tried to include them with a double-star pattern that does not. |

Each pattern is now listed in both forms. The `provider-test` rule already carried both forms for one
path, so this had been hit once before and not applied across the file.

Bringing `clientmock` under a rule surfaces its use of `github.com/google/uuid` — legitimate, for
assigning UUIDs to stored meshObjects — so the allow list now names it. **The 51 newly checked client
files produce no violations**: the client really was standard-library-only, it just was not being
verified.

Verified by compiling a file that violates each of the three rules and confirming depguard reports it.
That is the only way to tell an enforcing rule from an inert one.

## Commit 2 — the client move

- 112 files now import `github.com/meshcloud/meshstack-cli/client`. Nothing else about them changed,
  because the client kept its package path on the way over.
- Credential handling moves to the shared `pkg/login`. The four private `envKeyMeshstack` consts are
  gone; the variable names are exported from `pkg/login`, so both repositories define them once.
- `newProviderClient` merges the provider block attributes over `login.FromEnv()`, which keeps an
  explicitly configured attribute ahead of the environment. The per-attribute diagnostics stay
  per-attribute, because Terraform points them at the attribute that is missing, while
  `login.Authorization` returns a single error suited to a CLI.
- The `client` and `client-test` depguard rules travel with the client to meshstack-cli, where they
  still hold. That is what keeps external dependencies out of this provider's tree. The rule denying
  `client/internal` in provider tests is simply gone: it is in another module now, where Go's internal
  rule makes it unreachable outright.
- `AGENTS.md`, `DEVELOPMENT.md` and the `resource-development` / `modern-go` skills no longer point at
  a `client/` directory that is not here.

### One behaviour change

An empty `MESHSTACK_ENDPOINT` is now treated like an unset one and reports `Provider endpoint
missing.`. It used to be accepted, producing a client against an empty URL that failed later with a
worse message. `CHANGELOG.md` records it under FIXES.

### One dependency bump

testify moves from v1.11.1 to v1.12.0. Minimal version selection picks it up, because the client's
tests require it in the new module.

## Verification

- `task lint`: clean.
- **83 unit tests pass, no failures**, against the local meshstack-cli working copy.
- Acceptance tests against a live backend have **not** been run.
- Rebased onto current `main` (`8dbef43`, the GitHub platform type removal), which does not touch
  `client/`; the copy in meshstack-cli is byte-identical to `main`'s after normalising import paths.


grubmeshi:
Rebased onto current `main` and moved the changelog entries to a new `# v0.25.1` heading, anticipating that v0.25.0 ships first.

While moving them I found they had been sitting at the end of the **`# v0.24.4`** section, not `v0.25.0` as I first reported — v0.25.0 has only a BREAKING CHANGES block, and I anchored the original edit on a FEATURES bullet that belongs to v0.24.4. Since v0.24.4 is already tagged, that was worse than a stale heading. It is fixed now, and `v0.24.4` is back to its released content. Order reads `v0.25.1`, `v0.25.0`, `v0.24.4`, `v0.24.3`. No `Requires meshStack` line on `v0.25.1`, matching how the file treats releases that do not change the backend requirement.

The rebase also picked up two commits that had landed on `main`: your push of the depguard fix (so this branch is down to one commit), and `54b5d48 test: serialize acceptance tests that create restricted default tags`. That one touches `provider_test.go`, which this branch also edits, so I checked both changes survived — the `login.EnvKey*` references and the `pkg/login` import are intact.

Re-verified after the rebase: `task lint` clean, and the unit tests pass with no failures.

CI here stays red until the `replace` directive goes, which is the step after meshcloud/meshstack-cli#1 merges.

.agents/skills/modern-go/SKILL.md
```diff
@@ -59,9 +59,9 @@ than writing `any`-typed or reflection-based variants.
 
 | Type / func | File | Role |
 |---|---|---|
-| `MeshObjectClient[M any]`, `NewMeshObjectClient[M]`, `InferKind[M]()` | `client/internal/mesh_object_client.go` | Typed CRUD client per meshObject type |
```
grubmeshi: do not update instructions, but remove them as client things are now in CLI and should be added there (add a modern-go skill there as well)

grubmeshi:
Rebased onto current `main` (14 commits, `main` now at `e5d3075` with v0.25.0 released) and force pushed. Still
one commit.

## What the rebase needed

**Deleting `client/` is not enough on its own.** The rebase resolved the two modify/delete conflicts
(`client/client.go`, `client/landingzone.go`) as deletions, but `main` had meanwhile *added*
`client/client_test.go` and `client/mesh_info.go`. This commit does not delete files it never saw, so both
survived and `client/` came out of conflict resolution non-empty — with four files instead of none. Removing the
directory outright is the only resolution that means what the commit says.

**Two files on `main` still imported the old path.** `internal/clientmock/mock_mesh_info.go` and
`internal/provider/meshstack_instance_data_source.go` arrived with the meshStack instance data source and import
`terraform-provider-meshstack/client`, which no longer exists. Both now import
`github.com/meshcloud/meshstack-cli/client`, in the external group rather than the local one, since gci's
`localmodule` section no longer covers it. That is also how the `landingzone_upgrade.go` conflict is resolved —
`meshstack-cli/client` external, `internal/types/generic` local.

That brings the import sites from 112 to 114, and the commit message now says so.

## go.mod

`go mod tidy` was required after the rebase. testify goes to **v1.12.1** rather than the v1.12.0 the message
previously claimed, and `go.yaml.in/yaml/v3` comes along as testify's own transitive replacement for
`gopkg.in/yaml.v3`. Minimal version selection picks both up from the client's test requirements in the new
module. Message corrected.

The CHANGELOG entry stays under `v0.25.1`, which is now simply the next version rather than an anticipated one.

## Still local, still not pinned

The `replace github.com/meshcloud/meshstack-cli => ../meshstack-cli` directive is unchanged, so nothing is
pinned to a pseudo-version yet and meshstack-cli's `main` stays freely force-pushable. Dropping it and running
`go get github.com/meshcloud/meshstack-cli@main` is the step after
[meshstack-cli#1](https://github.com/meshcloud/meshstack-cli/pull/1) merges.

Note that meshstack-cli is now **Apache 2.0**, not MPL 2.0 — this provider stays MPL 2.0 and gains an
Apache-2.0-licensed dependency, which is compatible in that direction.

## Verification

Against the rebuilt meshstack-cli branch: `go build ./...` clean, the full unit suite passes, and
`golangci-lint run` reports 0 issues.

grubmeshi:
On CI: **Go Build** is red, and expected to stay red until meshstack-cli merges.

```
github.com/meshcloud/meshstack-cli@v0.0.0 (replaced by ../meshstack-cli):
reading ../meshstack-cli/go.mod: no such file or directory
```

The `replace` directive resolves against a sibling checkout that exists on a developer's machine and not on a
runner. This is not a regression from the rebase — every push on this branch has failed the same way since the
PR opened (`ab86cac`, `7a31acc`, now `e0c2c8c`), and the remaining jobs skip because they depend on Build.

It clears in one step, not by changing CI: once
[meshstack-cli#1](https://github.com/meshcloud/meshstack-cli/pull/1) is merged, drop the directive and run
`go get github.com/meshcloud/meshstack-cli@main`. Locally the same tree builds, tests and lints clean against
that branch, which is the check the directive exists to make possible.

grubmeshi:
All six jobs pass, including **Go Acceptance Test** (9m55s) — the first time this branch has had a full green
run, since every earlier push failed at Build on the `replace` directive and skipped everything downstream.

| Job | Result |
|---|---|
| Go Build | pass, 35s |
| Go Test | pass, 2m44s |
| Go Lint and Format Check | pass, 1m4s |
| Generate Terraform Provider Docs | pass, 21s |
| Shell Lint (CI scripts) | pass, 4s |
| Go Acceptance Test | **pass, 9m55s** |

The acceptance job is the one worth pointing at. It drives real Terraform against a real meshStack backend, so
it exercises the moved client over the wire — every request path, the JWT login exchange in
`client/internal/auth.go`, and the retry behaviour — rather than only checking that the provider compiles
against the new module. That is the evidence that the move changed no behaviour, and it was not obtainable at
all while the directive pointed at a sibling checkout.

PR #286, opened 2026-08-18
fix: honor MESHSTACK_SKIP_VERSION_CHECK before requesting /mesh/info

JohannesRudolph:
## The bug

In `checkMeshVersion` (`client/client.go`), the `MESHSTACK_SKIP_VERSION_CHECK == "true"` check sat
*inside* the `else if` branch that handles a version mismatch, so it was only reachable once
`GET /mesh/info` had already succeeded. Setting the flag therefore never skipped the request — it
only suppressed a version *mismatch*.

## The fix

Move the env check to the top of `checkMeshVersion`, so the flag short-circuits before the request is
built.

## Why it matters

`/mesh/info` is a `GET` issued on the *retrying* client (`client.New` wraps the HTTP client in
`internal.WithRetry` before calling `checkMeshVersion`, and `retryRoundTripper` retries all
idempotent methods, backing off on 503). So with the flag set, an unavailable meshStack did not fail
fast: every provider configure blocked for the client's full retry budget — `MaxRetries: 12` with the
`1+2+4+8+16+30*7` s backoff sequence, ~4 minutes — and then returned a hard error anyway, with no way
to opt out.

That is what the meshStack smoke tests hit: they set `MESHSTACK_SKIP_VERSION_CHECK=true` precisely to
decouple CI from this check, and it did not work. A dev-environment 503 window longer than the retry
budget (observed: 4–12 min) failed provider configuration for the entire test matrix at `init` time,
including in the destroy phase of tests that had already passed — leaving objects behind.

## Test

`client/client_test.go` covers both directions against a stub transport that always errors: with the
flag set, `checkMeshVersion` succeeds and the transport is never called at all; with the flag unset,
the request error surfaces. Verified to fail on the pre-fix code.

## Open question

Should a `/mesh/info` transport error be fatal at all, or a warning — a version *check* failing is
not the same as the API being unusable, and every subsequent call has its own retry/error handling?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ BREAKING CHANGES:
 - The deprecated `meshstack_tenant_v4` resource and data source have been removed, along with the `moved` state mover that migrated from it (the source schema it was built from no longer exists). Migrate to `meshstack_tenant` / `meshstack_tenants` by adding a `moved` block on v0.24.x — which still ships both the mover and the deprecated type — and applying it before upgrading to v0.25.0.
 - `meshstack_platform_types`: the `category` filter no longer accepts `GITHUB`. meshStack retired the dedicated GitHub platform type, and the platforms that had it are now `CUSTOM`. Change any `category = "GITHUB"` filter to `category = "CUSTOM"`.
 
+FIXES:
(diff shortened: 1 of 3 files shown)


grubmeshi (approved):
indeed, what a nice logic bug, thanks for fixing.

PR #288, opened 2026-08-20
feat: support adopting an existing identity provider on entra id integrations

malhussan:
## What this does

`meshstack_integration` gains `spec.config.entraid.idp_alias`. Set it to adopt an identity provider that already exists in a meshStack instance — typically one configured by hand — instead of having meshStack create one. Leave it out and meshStack generates an alias and reports it back. `meshstack_integrations` exposes it too.

The alias existed only in the internal API and the panel until now, so an existing identity provider could not be brought under Terraform management.

It also moves the Entra ID redirect URL into `status`, and adds the acceptance steps the Entra ID subtest was missing.

## Reading order

1. **`feat:`** — DTO field, resource and data source attributes, the immutability plan modifier, mock, generated docs, changelog.
2. **`test:`** — alias immutability, secret rotation, and adoption at create time.
3. **`fix!:`** — the redirect URL moves to `status.entraid.redirect_url`. Breaking.
4. **`docs:`** — the schema convention behind commit 3, recorded in the resource-development skill.

## The immutability decision

`idp_alias` is Optional + Computed, with `UseNonNullStateForUnknown` and a plan modifier that **rejects a change** rather than `RequiresReplace`.

meshStack deletes the Keycloak identity provider when an Entra ID integration is deleted, and records nothing about whether the alias was adopted or generated. A destroy-and-recreate would therefore delete a provider the customer configured themselves and registered a redirect URL for. Editing one string should not be able to do that, so the plan fails with an actionable error instead. Same pattern as `version_spec` in `building_block_definition_resource.go:401-415`; the framework ships no equivalent modifier, and a validator cannot see prior state.

## Why the redirect URL moved

`spec.config.entraid.redirect_url` → `status.entraid.redirect_url`.

meshStack derives the URL, so it belongs in a fully computed container. In `spec` it is unreachable under `mock_provider` and cannot be supplied by `override_*` either, which makes any module wiring that reads it untestable (#272). It was also Optional as well as Computed, and meshStack ignores a supplied value, so a configuration that set it failed the apply with `Provider produced inconsistent result after apply` — after creating the integration.

Nested per integration type, mirroring how `status.workload_identity_federation` nests per cloud, so further derived Entra ID values have somewhere to go.

## Verification

Unit suite (mock), `task lint` and the acceptance suite green; docs regenerated with `task generate`. All four commits build and vet on their own.

`05_entra_id_adopt_mock_only` is skipped against a real backend on purpose — a fixed alias cannot be re-adopted across runs, since meshStack rejects one another integration already uses.

## Backward compatibility

`idp_alias` is additive and optional, with no state upgrader. The redirect URL move is breaking for anyone reading the old path; the changelog names it.

Nothing in `meshstack-hub`, `internal-cloudfoundation`, `likvid-cloudfoundation` or `trial-cloudfoundation` configures an `entraid` integration, and every reference into `meshstack_integrations` there goes through `workload_identity_federation.*`. Verified by validating all four config shapes against the built provider.

### DIFF
--- a/.agents/skills/resource-development/SKILL.md
+++ b/.agents/skills/resource-development/SKILL.md
@@ -76,6 +76,16 @@ fields. Only refs mixing name *and* uuid (`MeshBuildingBlockV2TargetRef`) stay b
 
 Cross-cutting rules for the schema and its backing client, beyond the ref/DTO shape above:
 
+- **A computed value never sits inside a container configuration writes.** `metadata` and `spec` hold
+  inputs only; system-managed values belong top-level or in a fully computed container (`status`,
+  `ref`, `version_latest`). The reason is testability, not tidiness: OpenTofu's mock value generator
+  copies a config-written container verbatim and never descends into it, so a computed leaf under
+  `metadata`/`spec` is null under `mock_provider` and cannot be given a value by `override_*` either
+  — which makes any module wiring that reads it impossible to unit test. Depth is irrelevant; the
+  config-written ancestor is the only thing that disqualifies the subtree, and `Optional`+`Computed`
+  does not rescue it. When the API returns such a value inside `spec`, keep the client field
+  (`tfsdk:"-"`) and expose it from `status` via a local model struct — see the computed-only output
+  field pattern in `REFERENCE.md`. Tracked in #272.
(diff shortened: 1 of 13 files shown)


grubmeshi (approved):
LGTM, just some minor remarks in contect of #272

PR #289, opened 2026-08-20
feat: support optional Building Block Definition inputs

j0g3sc:
Exposes the optional-input flag added by meshcloud/meshfed-release#10701 on `meshstack_building_block_definition`, as `version_spec.inputs.*.is_optional`.

An operator can mark a definition input optional, so whoever fills the Building Block in may leave it blank and the implementation falls back to the default declared in its own code — a Terraform variable's `default`, a GitHub workflow input's `default`.

## What changed

- **Client DTO** — `IsOptional` on `MeshBuildingBlockDefinitionInput`.
- **Schema** — `is_optional` (Optional+Computed, default `false`), matching how `is_environment` and `updateable_by_consumer` are modelled.
- **Plan-time validation** — `validateOptionalInputs` in `ValidateConfig`.
- **Example + generated docs** — the `environment` input of `resource_01_terraform.tf` is now optional, which also exercises the field in the acceptance suite.
- **Tests** — `TestAccBuildingBlockDefinitionOptionalInputValidation` (8 cases, mock-client only, modelled on the manual-outputs validation test), `is_optional` added to the 9 exhaustive input state-check maps, and two content-hash fixtures.

## Notes for a reviewer

**`json:"isOptional,omitempty"` and no `currentHashVersion` bump.** `false` is the backend's default for an absent field, so a non-optional input serialises exactly as it did before the field existed. Every already-stored content hash stays byte-identical, which is what keeps a provider upgrade from re-running released Building Blocks — a v6 bump would have been the alternative. The new fixture `isOptional false hashes the same as example` pins that equality so a later change cannot break it silently. It also means a backend that does not know the field sees an unchanged payload for configs that do not use the feature.

**`MinMeshStackVersion` deliberately stays at 2026.34.0.** Unlike `meshstack_landingzone`'s `spec.restricted` (which the provider always sends and reads, so an older backend broke every apply), this field only reaches the backend for someone who opted in. Gating every user of the provider seemed to cost more than it protects. Instead the schema description and the changelog state the version requirement, and an older backend silently ignoring the field surfaces as Terraform's *"Provider produced inconsistent result after apply"* on `is_optional`. Happy to swap this for a version bump if you would rather have the hard gate.

**The four rules are mirrored client-side.** `ValidateConfig` rejects an optional input on the `manual` implementation, for an `assignment_type` other than `USER_INPUT`/`PLATFORM_OPERATOR_MANUAL_INPUT`, for type `BOOLEAN`, and alongside a `default_value` (including `sensitive.default_value`) — so these fail at plan instead of as a 400 mid-apply. Checks whose value is unknown at plan are skipped, so the backend stays the authority. The duplication is the trade-off: if the backend ever loosens a rule, this has to follow.

**Out of scope.** The `meshstack_building_block_definitions` data source exposes no `version_spec` at all, so there was nothing to add there. The instance side — read-only `isOptional`/`isSet` on `meshBuildingBlockV2` inputs and clearing an optional input with `"value": null`, from the same meshfed PR — is untouched and needs its own change to `meshstack_building_block`.

## Testing

`task lint` clean, full mock-mode suite (`go test ./...`) green. The acceptance suite has **not** been run against a real backend: the 01_terraform example now expects `is_optional` to round-trip, so `TestAccBuildingBlockDefinition` needs a meshStack built from the meshfed PR branch. That, and the exact meshStack version for the `2026.XX.X` TODO in the schema description, are open before this leaves draft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v0.25.1
+
+FEATURES:
+- `meshstack_building_block_definition`: new `version_spec.inputs.*.is_optional` argument. An optional input may be left blank by whoever fills the building block in; meshStack then sends no value and the implementation falls back to the default declared in its own code — a Terraform variable's `default`, a GitHub workflow input's `default`. It defaults to `false`, so nothing changes for an input you do not declare optional. Marking an input optional needs meshStack 2026.35.0 or later; an older backend silently ignores the field, which surfaces as Terraform's *"Provider produced inconsistent result after apply"* on `is_optional`. The provider mirrors the backend's four rules at plan time instead of letting them fail during apply: an optional input is rejected on the `manual` implementation, for any `assignment_type` other than `USER_INPUT` or `PLATFORM_OPERATOR_MANUAL_INPUT`, for type `BOOLEAN`, and together with a `default_value`. Existing versions keep their `content_hash`, so adding the attribute does not re-run any already-released building block.
+
 # v0.25.0
 
 Requires meshStack 2026.34.0 or later (previously 2026.30.0).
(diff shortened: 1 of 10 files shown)


j0g3sc:
Superseded by #290: this branch was renamed to `feature/CU-86c1hp0be_Optional-Building-Block-Inputs` so meshfed-release CI pairs it with the companion backend branch of the same name. GitHub closed this PR on the rename instead of retargeting it.

PR #291, opened 2026-08-24
feat: add an optional building block name template

grubmeshi:
## What

Exposes meshStack's new building block name template as `spec.display_name_template` on
`meshstack_building_block_definition`. It is a mustache-like template that names every new building block
after the values it was ordered with, for example `Project {{projectName}}`. Until now the definition's
`display_name` doubled as that template, so the raw placeholders also leaked into every place meshStack
shows a *definition* name.

Backend counterpart: meshcloud/meshfed-release#10762. **The acceptance tests here fail until that PR is in
`develop`**, because the field does not exist yet on the meshStack this repo's CI targets. That PR's
`tfAcceptanceTest` job tests both halves together, by branch name.

## Semantics

The attribute is a plain optional string and meshStack stores what it receives. No attribute, `null` and
`""` all mean the same thing to meshStack — name the building block after `display_name` — so a template
can be removed either by dropping the attribute or by setting it to `""`. Terraform tells those two apart,
so both round-trip through a refresh without leaving a diff behind.

## Tests

`TestAccBuildingBlockDefinition/15_display_name_template_lifecycle` walks set → change → empty → dropped.
The last two steps each carry a post-refresh `ExpectEmptyPlan`, which is what catches the perpetual-diff
trap that either representation of "no template" could otherwise cause.

`task lint` is clean, the unit-mode suite passes, and the paired `tfAcceptanceTest` against a local
meshStack carrying the backend PR passes (142 tests, 12 skipped). `task generate` refreshed the docs and
the `01_terraform` example.

## Not changed

`meshstack_building_block_definitions` deliberately exposes only the fields you select a definition by
(`display_name`, `target_type`), so the template is not added there.

`MinMeshStackVersion` stays at 2026.34.0. The gate is global and this is an additive optional field, which
`terraform-provider-compat` exempts from the version registry, so raising it would force an upgrade on
every user including those who never set the attribute. A configuration that does set it against a
meshStack that does not serve the field fails the apply on Terraform's consistency check; the CHANGELOG and
the attribute description both say so.

### DIFF
--- a/client/building_block_definition.go
+++ b/client/building_block_definition.go
@@ -24,6 +24,7 @@ type MeshBuildingBlockDefinitionMetadata struct {
 
 type MeshBuildingBlockDefinitionSpec struct {
 	DisplayName           string                `json:"displayName" tfsdk:"display_name"`
+	DisplayNameTemplate   *string               `json:"displayNameTemplate,omitempty" tfsdk:"display_name_template"`
 	TargetType            MeshBuildingBlockType `json:"targetType" tfsdk:"target_type"`
 	Description           string                `json:"description" tfsdk:"description"`
 	Readme                *string               `json:"readme,omitempty" tfsdk:"readme"`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ BREAKING CHANGES:
 
 FEATURES:
 - `meshstack_integration`: new `spec.config.entraid.idp_alias` argument adopts an identity provider that already exists in your meshStack, instead of having meshStack create one. Leave it out and meshStack generates an alias. The alias cannot be changed afterwards, and a plan that changes it fails. `meshstack_integrations` exposes it too.
+- `meshstack_building_block_definition`: new optional `spec.display_name_template` argument, a mustache-like template that names every new building block of the definition after the values it was ordered with, for example `Project {{projectName}}`. Without the argument, a new building block is named after `display_name`. This needs a meshStack that serves the field; an older one leaves it out of its response, so an apply that sets it fails Terraform's consistency check.
 
 # v0.25.0
 
--- a/docs/resources/building_block_definition.md
+++ b/docs/resources/building_block_definition.md
@@ -28,6 +28,7 @@ resource "meshstack_building_block_definition" "example_01_terraform" {
 
   spec = {
     display_name              = "Example Building Block"
+    display_name_template     = "Example Building Block {{ resource_name }}"                         # Optional: names each ordered building block after its inputs
     symbol                    = provider::meshstack::load_image_file("${path.module}/bb-symbol.png") # Optional
     description               = "An example building block definition"
     readme                    = "# Example Building Block\n\nThis is a comprehensive example showcasing all available attributes." # Optional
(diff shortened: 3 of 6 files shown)


client/building_block_definition.go
```diff
@@ -23,7 +23,10 @@ type MeshBuildingBlockDefinitionMetadata struct {
 }
 
 type MeshBuildingBlockDefinitionSpec struct {
-	DisplayName           string                `json:"displayName" tfsdk:"display_name"`
+	DisplayName string `json:"displayName" tfsdk:"display_name"`
```
vzaslavsky: d: For some reason, the alignment is gone.

AGENTS.md
```diff
@@ -93,6 +93,7 @@ require them.
 | `scratch-config` | Standalone repro/debug/prototype against any meshStack you own |
 | `changelog-management` | Pick the next version, maintain `CHANGELOG.md` |
 | `github-ci` | GitHub Actions conventions, action SHA-pinning |
+| `pr-review-comments` | Work through a PR's review comments one at a time, replying and resolving |
```
vzaslavsky: d: Something unrelated, I guess.

internal/clientmock/mock_building_block_definition.go
```diff
@@ -61,13 +61,29 @@ func (m meshBuildingBlockDefinitionClient) Create(_ context.Context, definition
...
+func updatedNameTemplate(stored, requested *string) *string {
+	switch {
+	case requested == nil:
+		return stored
+	case *requested == "":
+		return nil
+	default:
+		return requested
+	}
+}
```
vzaslavsky: d: I don't think I understand why we need it here. I thought, we're later calling the API with the values from the Terraform file. Since we handle this in the API, shouldn't we just pass the values through?

vzaslavsky (approved):
Looks alright!

Depending on how it goes in the companion PR, we might want to adjust the docs and the model a bit. The only question I have in this one is: whether we really need to copy the new field handling here.

AGENTS.md
grubmeshi: yes, will remove it

internal/clientmock/mock_building_block_definition.go
grubmeshi: it's a bit tricky with "inconsistent result after apply" in terraform which this little snippet addresses. the plan would be to make that required once we've gracefully introduced the optional attribute here.

vzaslavsky: Ah, this is what I suspected. Thanks for confirming that!

grubmeshi: ah, now that i've though more, let me steer it into the simply direction of having it truely nullable in tf resource and api and not having the backend treat null as no change, as we simply take whatever makes sens in the frontend already (name as fallback, otherwise we use non-null/non-empty, making empty stirng semantically equal to null value (terraform still makes destiction here)

vzaslavsky: Yeah, if that produces no "inconsistent ..." thingie, then it would be a much cleaner solution 👍

client/building_block_definition.go
grubmeshi: Fixed. The comment inserted mid-struct is what split the alignment group, and it described the keep-on-absent behaviour, which is gone — so I deleted it and gofmt realigned the block.

AGENTS.md
grubmeshi: Removed.

internal/clientmock/mock_building_block_definition.go
grubmeshi: Done, in the direction of your last comment. `updatedNameTemplate` in the mock, `keepClearedNameTemplate` and the `Read` hook are all gone; the mock passes the spec through because the API does too. No attribute, `null` and `""` all mean the same thing to meshStack, Terraform tells the last two apart, and both round-trip.

Confirmed against a real backend: the paired `tfAcceptanceTest` in meshcloud/meshfed-release#10762 passes, and the lifecycle test's set → change → empty → dropped steps each end with a post-refresh `ExpectEmptyPlan`.

PR #292, opened 2026-08-26
feat: expose building block definition approvals and drift schedule

nroi:
The meshBuildingBlockDefinition meshObject API now carries the meshPanel
Policies tab in spec.approvals and spec.scheduleConfig (meshfed-release
#10473), so meshstack_building_block_definition exposes both as arguments.
They default to no approval gate and no schedule, which is what meshStack
stores for a new definition.

meshStack validates both against the implementation type of the
definition's latest version, and it rejects a version implementation-type
change while the stored policies are incompatible with the new type. So a
policy the current type cannot honour has to be written after version_spec,
and one the new type cannot honour has to be gone before it. Neutral
policies satisfy every implementation type, so Create writes the definition
neutral, writes the version, then writes the planned policies, and Update
passes through neutral in between when the implementation type changes.
That makes a combined implementation-type and policy change apply in one
step.

Two object validators repeat meshStack's rules at plan time, so an
unsupported combination fails before anything is written. The mock client
learns the same rules, so a unit-test run catches an ordering regression
without a live backend.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


internal/provider/schema_utils.go
```diff
@@ -29,6 +30,32 @@ func nestedObjectToObjectType(nested schema.NestedAttributeObject) types.ObjectT
 	return types.ObjectType{AttrTypes: attrTypes}
 }
 
+// attributeDefaultsObject builds an object whose every attribute holds that attribute's own schema
+// default. Use it for the Default of an Optional+Computed SingleNestedAttribute, so the container's
+// default cannot drift from its leaves': omitting the whole block then yields exactly what declaring
+// it as an empty block does. It panics on an attribute kind it cannot read a default from, or on one
+// without a default - both are provider bugs that surface on every provider start.
+func attributeDefaultsObject(ctx context.Context, attributes map[string]schema.Attribute) types.Object {
```
nroi: i: I'm not very happy with this approach tbh, it feels like a homemade solution for a problem that should have a better solution via a 3rd party library or the terraform plugin framework, but I let claude run some analysis and it did not come up with a better solution. See the analysis from claude in the response.

nroi: **[AI-generated content]**

Analysis of whether a library or framework feature can replace `attributeDefaultsObject`. Short version: it cannot. The framework does not have this feature, no third-party package can provide it, and the other providers I surveyed all solve it by hand — most of them worse.

### The gap is a known, open upstream request

[terraform-plugin-framework#777, "Consider Object Nested Attribute Default Implementation"](https://github.com/hashicorp/terraform-plugin-framework/issues/777), open since June 2023, describes exactly what this helper does:

> When working with nested attributes, it may be desirable to automatically have an object default based on all the nested attribute default values for when the object is null. Today, this is a manual process.

bflad's proposal is to add `AttributeTypes` and `NestedAttributes` to `defaults.ObjectRequest` and ship an `objectdefault.NestedAttributeDefaults()`. Nobody has picked it up. Two related issues are also open: [#726](https://github.com/hashicorp/terraform-plugin-framework/issues/726) (defaults in nested attributes) and [#939](https://github.com/hashicorp/terraform-plugin-framework/issues/939), which asks for the documentation to cover this case at all. The current docs page only says nested attribute defaults "are applied afterwards", which is misleading for the null-parent case.

### Why the framework cannot fill the leaves itself

I read the mechanism in terraform-plugin-framework v1.19.0 to confirm the helper is not redundant with framework behaviour.

`Data.TransformDefaults` (`internal/fwschemadata/data_default.go`) applies every default by walking the plan value with `tftypes.Transform`. Two properties of that walk decide the outcome:

1. `transformUnderlying` starts with an explicit early return — "If the Value is null or unknown, there is nothing to descend." When the user omits the whole block, the plan value at that path is a null object, so the leaf paths are never visited and their `Default` never runs.
2. `transform` calls `transformUnderlying` first and only then calls the callback on the node itself. The traversal is post-order: children before parent. So even when the object default replaces the null object with a known one, the leaves have already been passed over. No traversal order exists in which the framework fills the leaves after the parent default fires.

That is why the proposed upstream fix has to pass the nested schema into the object default, rather than change the walk.

### Why no library can provide it either

A custom `defaults.Object` implementation cannot read the leaf defaults at plan time. `defaults.ObjectRequest` has exactly one field, `Path` — no config, no schema. Adding that is precisely what #777 asks for.

The other route is reading the defaults off the schema at construction time, which is what this helper does. It has to type-switch on the concrete `schema.BoolAttribute` / `schema.StringAttribute` structs because the framework's own accessor interfaces (`AttributeWithBoolDefaultValue` and friends) live in `internal/fwschema` and are not importable. So an external package cannot write this generically for all attribute kinds either — it would end up with the same type switch and the same `default` branch. That is also the real reason for the `panic`: it marks the boundary of what is reachable from outside the framework, not a shortcut.

### What other providers actually do

I searched GitHub for `objectdefault.StaticValue` (208 files) and read the interesting ones.

- **terraform-provider-aws** — `resource/schema/booldefault` appears in 35 files, `resource/schema/objectdefault` in **zero**. The largest framework-based provider never puts a default on a nested object. Nested containers stay plain `Optional`, and the flatten/expand layer normalizes.
- **cloudflare** (`internal/services/worker/schema.go`) — writes the object default as a typed model literal: `objectdefault.StaticValue(customfield.NewObjectMust(ctx, &WorkerObservabilityTracesModel{Enabled: types.BoolValue(false), HeadSamplingRate: types.Float64Value(1), ...}).ObjectValue)`. Their leaves carry no `Default:` at all; they use per-attribute plan modifiers such as `PropagationPolicyDefault()` instead. The duplication moves into the modifiers rather than disappearing.
- **coralogix** (`internal/provider/alerts/alert_schema/common.go`) — writes `objectdefault.StaticValue(types.ObjectValueMust(UndetectedValuesManagementAttr(), map[string]attr.Value{...}))` **and** declares `Default:` on the leaves. This is exactly the drift hazard the doc comment warns about, live in a shipping provider.
- **elastic** (`internal/fleet/integration_policy/schema.go`) — sidesteps it with `objectdefault.StaticValue(basetypes.NewObjectNull(...))`, a null object default, so there is nothing to keep in sync.
- **descope** (`internal/models/attrs/objattr/object.go`) — built a full generic object-attribute abstraction layer, and its `Default[T]` still takes a handwritten value.

So the field standard is either "duplicate the values by hand" or "do not default the object". None of them derives the object default from the leaves. This helper is the only version of it I found that closes the drift.

### On the `panic`

It runs inside `Schema()`, so an unsupported attribute kind crashes the provider on the first call — in every mock unit test and in CI. It cannot reach a user without failing the build first. It also matches what this repo already does for the same class of bug: `internal/provider/tenant_upgrade.go:45` panics with `"tenant spec attribute is not a SingleNestedAttribute"`, and `landingzone_upgrade.go:36` and `building_block_upgrade.go:39` do the same. There are 13 uses of `types.…ValueMust` in non-test code, including `emptySetDefault` and `tagsAttribute` in this very file; `ObjectValueMust` on the last line of this helper already panics on a type mismatch.

The one alternative worth naming: `Schema(ctx, req, resp)` has `resp.Diagnostics`, so the helper could take a `*diag.Diagnostics` and report an error instead of panicking. I would not do it — the caller would then have to handle a meaningless zero-value object, and the failure is a provider bug of compile-time shape, not a runtime condition.

### Options, if we want to change something

1. **Leave it as is**, and add a unit test asserting that the derived object equals the leaf defaults for both call sites, plus one asserting the `default` branch panics for an unsupported kind. That gives the invariant a name in CI instead of relying on the provider-start crash.
2. **Extend the type switch to the remaining kinds** (`Int64`, `Float64`, `List`, `Map`, `Set`, `Object`) so the panic becomes unreachable for anything we would nest. Mechanical, roughly 40 lines, but dead code until a second shape needs it.
3. **Contribute the helper upstream** as the `objectdefault.NestedAttributeDefaults()` that #777 asks for. Inside the framework the internal interfaces are reachable, so it can be written for all attribute kinds without a type switch on concrete structs and would not need to panic. This is the only route that removes our code rather than moving it.

client/building_block_definition.go
```diff
@@ -16,28 +16,91 @@ var (
...
-	UseInLandingZonesOnly bool                  `json:"useInLandingZonesOnly" tfsdk:"use_in_landing_zones_only"`
-	SupportURL            *string               `json:"supportUrl,omitempty" tfsdk:"support_url"`
-	DocumentationURL      *string               `json:"documentationUrl,omitempty" tfsdk:"documentation_url"`
+	DisplayName           string                               `json:"displayName" tfsdk:"display_name"`
+	DisplayNameTemplate   *string                              `json:"displayNameTemplate,omitempty" tfsdk:"display_name_template"`
+	TargetType            MeshBuildingBlockType                `json:"targetType" tfsdk:"target_type"`
+	Description           string                               `json:"description" tfsdk:"description"`
+	Readme                *string                              `json:"readme,omitempty" tfsdk:"readme"`
+	RunTransparency       bool                                 `json:"runTransparency" tfsdk:"run_transparency"`
+	ApprovalPolicies      MeshBuildingBlockDefinitionApprovals `json:"approvalPolicies" tfsdk:"approval_policies"`
```
grubmeshi: f: align?
```suggestion
	ApprovalPolicies      MeshBuildingBlockDefinitionApprovalPolicies `json:"approvalPolicies" tfsdk:"approval_policies"`
```

nroi: fixed

internal/provider/schema_utils.go
grubmeshi: Ah I remember @henryde having a similar struggle. What you can do though is to modify the DTO maybe on Read  and supply a non-null default if the backend does not fill it? maybe that's sth worth trying.

internal/provider/building_block_definition_resource_test.go
```diff
@@ -1448,6 +1484,238 @@ func checkAzureDevopsPipelineImplementation() knownvalue.Check {
...
+	))
+}
+
+// TestAccBuildingBlockDefinitionPolicies covers changing spec.approval_policies / spec.schedule
+// together with the version's implementation type in one apply. meshStack validates both policies
+// against the implementation type of the definition's latest version, and rejects an implementation
+// change while the stored policies are incompatible with the new type. Neither direction can be
+// applied with a single definition write, so this asserts the resource orders the writes: it passes
+// through neutral policies, writes the version, then writes the planned policies.
+func TestAccBuildingBlockDefinitionPolicies(t *testing.T) {
```
grubmeshi: f: see if you can merge this into existing acc tests (subtests above) instead of coming up with another top-level test

internal/clientmock/mock_building_block_definition.go
```diff
@@ -60,12 +66,66 @@ func (m meshBuildingBlockDefinitionClient) Create(_ context.Context, definition
...
+	}
+	if err := m.validatePolicies(uuid, existing, definition.Spec); err != nil {
+		return nil, err
+	}
+	existing.Spec = definition.Spec
+	existing.Metadata.Tags = definition.Metadata.Tags
+	return existing, nil
+}
+
+// validatePolicies mirrors the backend's rule that the approvals and the drift schedule must be
```
grubmeshi: f: unless absolutely necessary, do not replicate backend logic in mock too much. I think simply "echoing" in mock mode is perferctly find, and we do not need extended validation in tf provider unless easy to achieve (the backend will tell us on apply with a meaning error description instead)

grubmeshi (approved):
LGTM, trying to keep the code change small and thus challenging it a bit

internal/clientmock/mock_building_block_definition.go
nroi: fixed

internal/provider/building_block_definition_resource_test.go
nroi: fixed

internal/provider/schema_utils.go
nroi: >  What you can do though is to modify the DTO maybe on Read and supply a non-null default if the backend does not fill it?

Not sure if I understood the idea correctly, but `Read` does not fire if we have for example a fresh `main.tf` without any prior state and now run a `tofu plan` for a resource that doesn't yet exist remotely. So, whatever we write in `Read` won't have any impact what the `tofu plan` shows us for `spec.approval_policies`.

